PR #2, opened 2019-12-17
Docker modifications

bsb808:
This is a little different workflow for docker.

* Assumes that the container will mount source files on the host.  This means we just need one copy of the source instead of one on the host and one in the container.
* The source is built (catkin_make) when the container is run, not when it is generated.  A bit easier for development, but not for pipelines.

Also cleaned up some of the VRX bits and added more dependencies to the apt installs.

See https://github.com/Field-Robotics-Lab/DAVE/wiki/DAVE-with-Docker

### DIFF
--- a/docker/uuvsim/Dockerfile
+++ b/docker/uuvsim/Dockerfile
@@ -67,6 +67,18 @@ RUN /bin/sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) 
     ros-${DIST}-velodyne-simulator \
     ros-${DIST}-rqt \
     ros-${DIST}-rqt-common-plugins \
+		ros-${DIST}-effort-controllers \
+		ros-${DIST}-control-toolbox \
+ 		ros-${DIST}-controller-manager \
+		ros-${DIST}-ros-control \
+		python-scipy \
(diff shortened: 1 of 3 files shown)


docker/uuvsim/Dockerfile
```diff
@@ -102,15 +114,15 @@ RUN mkdir -p uuv_ws/src/uuv_simulator
 
 # Copy the VRX repository from the local file system
 # We can't use the USER:GROUP variables until Docker adds support to --chown
-COPY --chown=developer:developer . uuv_ws/src/uuv_simulator
+#COPY --chown=developer:developer . uuv_ws/src/uuv_simulator
 
 # Compile the VRX project.
-RUN /bin/bash -c ". /opt/ros/${DIST}/setup.bash && cd uuv_ws && catkin_make"
+#RUN /bin/bash -c ". /opt/ros/${DIST}/setup.bash && cd uuv_ws && catkin_make"
```
M1chaelM: This avoids the need to pull in files from the host system in order to build the project in the image. A drawback is that the image won't contain a pre-built version of the system; the user will have to run catkin_make first before it can be used. I think this is good for our use-case but if we ever want to put the image in a repository so non-developers can just run it we will probably need to change this.

PR #3, opened 2019-12-23
Adding a rectangular bar to pick up

M1chaelM:
This pull request adds a simple rectangular bar that can be grasped by the Oberon 7 arm. A cylinder is also a possibility, but much more difficult to grasp properly.

Currently, in order for this to work properly, the included dave_ocean.world file must first be copied to the uuv_simulator/uuv_gazebo_worlds/worlds/dave_ocean.world. I created a pull request because I'm not sure this setup is correct, so I'd appreciate a review.


BruceMty:
Rather than making a new world file, lets instead use spawn_model to load the bar.

PR #11, opened 2020-05-09
DVL Model and Plugin Update

dtdavi1:
Updated dvl_gazebo package to include URDF (Xacro) models for performer-specific DVLs using the UUV Simulator DVL plugin.  

Added SDF and URDF models and examples implementing the WHOI DSL environment plugins.

The DVL tutorial on the wiki page (https://github.com/Field-Robotics-Lab/dave/wiki/whn_dvl_examples), which corresponds to the package README.md file, has been updated to include instructions and documentation for the new UUV Simulator models and the WHOI DSL environment models.

### DIFF
--- a/dvl_gazebo/README.md
+++ b/dvl_gazebo/README.md
@@ -6,26 +6,40 @@ https://www.apache.org/licenses/LICENSE-2.0
 Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the License for the specific language governing permissions and limitations under the License.
 
 ## Overview
-The ROS `dvl_gazebo` package provides SDF, URDF, and Xacro files for utilization, testing, and evaluation of the UUV Simulator Doppler Velocity ogger (DVL) plugin.  The included models and plugin parameters roughly align with the specifications of the Teledyne Workhorse Navigator 600.  Model and world files can be used as is for testing of the DVL model.  The Xacro file can be used to add a DVL to an existing robot model.
+The ROS `dvl_gazebo` package provides SDF, URDF, and Xacro files for utilization, testing, and evaluation of the UUV Simulator  and Woods Hole Oceanographic Institute (WHOI) Deep Submergence Laboratory (DSL) environment ROS Doppler Velocity Logger (DVL) plugins.  The included exemplar models and plugin parameters utilized by the fully-functional launch files described in this tutorial roughly align with the specifications of the Teledyne Workhorse Navigator 600.  Model and world files can be used as is for testing of the DVL model.  The Xacro files can be used to add a DVL to an existing robot model.
 
-The package contains two launch files that can be run independently (both covered in more detail below) that can be executed with the following commands:
+Xacro files are provided for the following DVLs are provided in addition to the WHN 600 of the executable examples:
+
+- **Nortek DVL 500-300m**
+- **Nortek DVL 500-6000m**
+- **Nortek DVL 1000-300m**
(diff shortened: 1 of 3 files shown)


bsb808:
Fantastic.  I'm a little backlogged with class and the sonar plugin.  @BruceMty would you have time to review?

BruceMty:
I can look at this tomorrow.

bsb808:
Running the tests...

Both of the WHN examples generate the following warnings...
`roslaunch dvl_gazebo uuvsim_teledyne_whn_standalone.launch`

```
Warning [parser.cc:626] Can not find the XML attribute 'version' in sdf XML tag for model: NED frame. Please specify the SDF protocol supported in the model configuration file. The first sdf tag in the config file will be used 
[INFO] [1589318904.850189, 0.000000]: Calling service /gazebo/spawn_urdf_model
Warning [parser_urdf.cc:1119] multiple inconsistent <gravity> exists due to fixed joint reduction overwriting previous value [true] with [false].
Warning [parser_urdf.cc:1119] multiple inconsistent <gravity> exists due to fixed joint reduction overwriting previous value [true] with [false].
Warning [parser_urdf.cc:1119] multiple inconsistent <gravity> exists due to fixed joint reduction overwriting previous value [true] with [false].
Warning [parser_urdf.cc:1119] multiple inconsistent <gravity> exists due to fixed joint reduction overwriting previous value [true] with [false].
```

Perhaps this is expected since...
```
NOTE for all models and macros: Gravity has been disabled for all of these models for the time being so that they can be incorporated without additional plugins and controllers.
```

BruceMty:
Wiki questioins, ref. https://github.com/Field-Robotics-Lab/dave/wiki/whn_dvl_examples#uuv-simulator-plugin-utilization:

Launch files are not present.  Were they renamed?

Does "choosing to "follow" the teledyne_whn model" mean to keep the DVL in the Gazebo GUI view by moving the mouse?

Running "roslaunch dvl_gazebo teledyne_whn_standalone.launch":
1) Only 1 of 4 ray sets is blue.  Is this expected?
2) Can we resolve two warnings: a) XML attribue 'version' and b) multiple inconsistent <gravity>?

bsb808:
> Does "choosing to "follow" the teledyne_whn model" mean to keep the DVL in the Gazebo GUI view by moving the mouse?

In Gazebo - you can locate the model on the left, then right-click on it and select "Follow" so the camera follows the model.

bsb808:
I was able to run the WHOI examples without issue.

Great job on the writeup / notes.

It seems we should schedule some time to discuss where to go from here.  I suspect that the WHOI model is the preferred starting point and that we can start adding functionality from there.

bsb808:
I approved - assuming that @dtdavi1 will resolve the warnings comment above.

Recommend working with @BruceMty to sort out the issues he discusses, but appears good to go from my location.

dtdavi1:
The missing "version" warning comes from the NED frame model that is included in the world file.  Both the world file itself and the included NED file are in the UUV Simulator codebase--I can make the warning go away, but the "fix" would be limited to our local fork.

The gravity warnings are a byproduct of the gravity being disabled for the links for now (since the "robot" isn't powered, the whole thing just falls like a rock if gravity is enabled).

dtdavi1:
@BruceMty, I left a separate comment about the warnings (we're stuck with the gravity ones for now, and the version one comes from the UUV Simulator codebase--neither actually means anything functionally).

I'm not sure about the ray color--they should all be blue, but can render differently depending on the viewing angle and light position.

The names did change in the dvl_dev branch (I moved them when I added the WHOI sim stuff) but the changes were already reflected in the wiki tutorial.  I've completed the merge, so if you update from master, then everything should work per the tutorial.

PR #14, opened 2020-05-14
Bop panel demo

Youssef-Khaky:
**Pull request overview.**
Added a new launch file that spawns the uuv in front of a bop panel. The purpose of this is to demonstrate the capability of the simulation and users can also use this to get used to controlling the manipulator.  

**Details**  

- `valve_manipulation_demo.launch`  

The launch file does the same thing as the `uuv_dave.launch` and also includes a launch file called `rexrov_oberon_arms_demo.launch` that spawns the uuv model with the position hold functionality.

- `uuv_dave_ocean_waves.world`  

Some light elements were added to the world to the panel can be visible.  
 

**Associated wiki:**  
https://github.com/Field-Robotics-Lab/dave/wiki/BOP-Panel-Manipulation-Mission.  
Comments
I am not sure if there is something wrong with the manipulator, but it is quite difficult to get it to hold tightly on the knob dial.  
It ends up looking something like this.
manipulator


bsb808:
Great.  I've got a full day, but will try to take a look tonight and/or tomorrow.

uuv_dave/launch/valve_manipulation_demo.launch
```diff
@@ -0,0 +1,83 @@
+<!--
```
bsb808, line 1: First line should be `<?xml version="1.0"?>`

uuv_dave/launch/valve_manipulation_demo.launch
```diff
@@ -0,0 +1,83 @@
...
+        <arg name="world_name" value="$(arg world_name)"/>
+        <arg name="paused" value="$(arg paused)"/>
+        <arg name="use_sim_time" value="true"/>
+        <arg name="gui" value="$(arg gui)"/>
+        <arg name="headless" value="false"/>
+        <arg name="debug" value="false"/>
+        <arg name="verbose" value="true"/>
+    </include>
+
+    <include file="$(find uuv_gazebo)/launch/rexrov_demos/rexrov_oberon_arms_demo.launch">
```
bsb808, line 22: This launch file contains a lot of the functionality.  I would propose that you refactor it so it is a bit more modular and reusable.  We can use the existing launch file as a working example and split the functionality into a set of launch files in the dave repository (or put pieces directly in this launch file)  that...

1. Configure and launch the thruster manager and controllers.
2. Configure and launch the teleoperation and launch the joint control for manipulation
  * It would be nice if there was a command line parameter to allow for swtiching between the xbox and logitech configurations.
3. Launch rviz directly from this launch file - group with the image viewer.

Youssef-Khaky: As I was doing this I realized there was already a lot of modularized code available. Should be much clearer now.

Youssef-Khaky:
Thanks for the detailed feedback. I have implemented the changes requested.

bsb808:
Looks great.  I'm still futzing with the gamepad controls, but I think that is a config issue on my end.

Could you please resolve the the conversation elements above?  This is a good way to account of addressing each issue in the review.  If you were able to address, go ahead and resolve.  If there are further questions you can add to the individual conversation thread.  This way we make sure we address each item.

Otherwise good to go.  Ready to merge.

PR #15, opened 2020-05-19
Adding sonar design sketch

bsb808:
Simply adding graphml sketch and png.  Sketch generated with yEd.


BruceMty:
Maybe we should not have a docs directory in this repository.  How about if we have a directory in the Wiki repository instead.  This way it is easier to link to graphs from the Wiki and adding content does not need peer review.  Also, please add a Makefile for building the graph, see Makefile in dave.wiki/drawings.

bsb808:
I put the drawings there (dave.wiki) first, but it it is tough to browse them via the web.  Is there a way to see the drawings directly in a web browser?  A view similar to this...
https://github.com/Field-Robotics-Lab/dave/blob/docs/docs/sonar_dev_arch.png

BruceMty:
I see.  For direct browser navigation, it needs to be on the repository side.  I'm fine with a separate "dave_doc" repository or as is.

PR #16, opened 2020-05-21
Manipulator force feedback

Youssef-Khaky:
**Overview**
Added FT sensors to all arm joints and the gripper joints. (From Azimuth joint up to wrist, no FT sensors in the finger joints)  
Related wiki can be found [here](https://github.com/Field-Robotics-Lab/dave/wiki/Manipulator-Force-Feedback).
Video demonstrating functionality can be found [here](https://vimeo.com/422486993).
**Details**  

- Created a world with a springy place in it called `uuv_dave_ocean_waves_spring.world`
- new launch file `uuv_manipulator_spring_demo.launch` to launch the uuv in that world
- copied some urdf files from the manipulator repo to add the FT sensor functionality. These files are:

1. oberon7_default.xacro (paths were changed to open the serial arm from the uuv dave)
2. serial_arm.gazebo.xacro (ft sensors were added here)
3. serial_arm.xacro (ft sensor macro was added to this file)

- The `rexrov_oberon7.xacro` file was edited to open the new urdf files in the uuv_dave repo and not the manipulator arm.

### DIFF
--- a/uuv_dave/launch/uuv_manipulator_spring_demo.launch
+++ b/uuv_dave/launch/uuv_manipulator_spring_demo.launch
@@ -2,7 +2,7 @@
 <launch>
     <arg name="gui" default="true"/>
     <arg name="paused" default="false"/>
-    <arg name="world_name" default="$(find uuv_dave)/worlds/uuv_dave_ocean_waves_spring.world"/>
+    <arg name="world_name" default="$(find uuv_dave)/worlds/uuv_dave_ocean_waves_simple.world"/>
 
  
     <!-- use Gazebo's empty_world.launch with uuv_dave_ocean_waves.world -->
@@ -19,6 +19,10 @@
     <!-- use ned frame north east down -->
     <include file="$(find uuv_assistants)/launch/publish_world_ned_frame.launch"/>
 
+    <!-- springy plate -->
+    <node name="spawn_springy_plate" pkg="gazebo_ros" type="spawn_model"
+          args="-file $(find uuv_dave)/models/springy_plate/model.sdf -sdf -model springy_plate">
+    </node>
 
     <!-- rexrov robot with oberon7 arm -->
     <include file="$(find uuv_dave)/launch/upload_rexrov_oberon7.launch">
--- a/uuv_dave/models/springy_plate/model.config
+++ b/uuv_dave/models/springy_plate/model.config
@@ -1,0 +1,16 @@
+<?xml version='1.0'?>
+
+<model>
+  <name>springy_plate</name>
+  <version>1.0</version>
+  <sdf >model.sdf</sdf>
+
+  <author>
+    <name>Youssef Khaky</name>
+    <email>khakyya@gmail.ca</email>
+  </author>
+
+  <description>
+    A plate that decompresses agaienst a spring
+  </description>
+</model>
--- a/uuv_dave/models/springy_plate/model.sdf
+++ b/uuv_dave/models/springy_plate/model.sdf
@@ -1,76 +1,5 @@
-<?xml version="1.0" ?>
-<!-- Copyright (c) 2016 The UUV Simulator Authors.
-     All rights reserved.
-
-     Licensed under the Apache License, Version 2.0 (the "License");
-     you may not use this file except in compliance with the License.
-     You may obtain a copy of the License at
-
-       http://www.apache.org/licenses/LICENSE-2.0
-
-     Unless required by applicable law or agreed to in writing, software
-     distributed under the License is distributed on an "AS IS" BASIS,
-     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-     See the License for the specific language governing permissions and
-     limitations under the License.
--->
-<!-- Modified for different Gazebo GUI camera angle -->
-
-<sdf version="1.4">
-  <world name="oceans_waves">
-    <physics name="default_physics" default="true" type="ode">
-      <max_step_size>0.002</max_step_size>
-      <real_time_factor>1</real_time_factor>
-      <real_time_update_rate>500</real_time_update_rate>
-    </physics>
-    <scene>
-      <ambient>0.01 0.01 0.01 1.0</ambient>
-      <sky>
-        <clouds>
-          <speed>12</speed>
-        </clouds>
-      </sky>
-      <shadows>1</shadows>
-    </scene>
-
-    <!-- Origin placed somewhere in the middle of the North Sea  -->
-    <spherical_coordinates>
-      <latitude_deg>56.71897669633431</latitude_deg>
-      <longitude_deg>3.515625</longitude_deg>
-    </spherical_coordinates>
-
-    <!-- Global light source -->
-    <light type="directional" name="sun1">
-      <pose>50 0 150 0 0 0</pose>
-      <diffuse>1 1 1 1</diffuse>
-      <specular>.1 .1 .1 1</specular>
-      <direction>0.3 0.3 -1</direction>
-      <cast_shadows>false</cast_shadows>
-    </light>
-
-    <!-- Global light source -->
-    <light type="directional" name="sun_diffuse">
-      <pose>-50 0 -150 0 0 0</pose>
-      <diffuse>0.6 0.6 0.6 1</diffuse>
-      <specular>0 0 0 1</specular>
-      <direction>-0.3 -0.3 -1</direction>
-      <cast_shadows>false</cast_shadows>
-    </light>
-
-    <!-- Virtual NED frame -->
-    <include>
-      <uri>model://ned_frame</uri>
-      <pose>0 0 0 0 0 0</pose>
-    </include>
-
-    <!-- Bounding box with sea surface -->
-    <include>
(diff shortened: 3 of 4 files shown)


uuv_dave/urdf/parallel_gripper.gazebo.xacro
```diff
@@ -0,0 +1,62 @@
...
+                <frame>child</frame>
+                <measure_direction>child_to_parent</measure_direction>
+                <always_on>true</always_on>
+                <update_rate>30</update_rate>
+                </force_torque>
+            </sensor>
+        </gazebo>
+
+        <gazebo reference="${namespace}/finger_left_joint">
+            <sensor name="arm_finger_left_joint" type="force_torque">
```
Youssef-Khaky, line 17: @bsb808 I just want to make sure I am understanding this right. This sensor tag is completely independent of ROS. All it does is publish wrench messages to a gazebo topic. So what I need to do to get this to a ROS topic is to write a ROS-aware plugin that reads those force_toruqe messages from the gazebo topic and republish them to a ROS topic.  
Is my understanding correct?

bsb808: @Youssef-Khaky Yes, you have explained it well.  This is a typical design pattern for Gazebo/ROS.  Working with both can feel a bit like loving in parallel universes simultaneously.  

I imagine you saw this tutorial - http://gazebosim.org/tutorials?tut=force_torque_sensor&cat=sensors   The `gz topic` command is a good introspection tool to see the traffic, but not as well documented as `rostopic`

It looks like the `force_torque` gazebo sensor publishes a WrenchStamped gazebo message (protobuf) which you can see here - https://github.com/osrf/gazebo/tree/gazebo11/gazebo/msgs
I also typically take a quick look at the sensor code to better understand how it works - https://github.com/osrf/gazebo/tree/gazebo11/gazebo/sensors

There are at least two ROS plugins that do some of what you want.
 
1. The FT ROS plugin: https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_ft_sensor.cpp .   This includes some noise.   Here is an example of using the sensor that looks reasonable: https://github.com/CentroEPiaggio/force-torque-sensor/blob/master/model/ft_sensor.gazebo.xacro  
2. F3D ground truth plugin: http://gazebosim.org/tutorials?tut=ros_gzplugins  and https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_f3d.cpp  which doesn't include noise.

You might take a look at the code for 1 and 2 to verify that they are doing the same thing with the exception of adding noise.

One thing I find useful in planning this type of thing is to draw a quick "boxes and arrows" sketch of what is going on.  Here is an example from the sonar effort:  https://github.com/Field-Robotics-Lab/dave/blob/master/docs/sonar_dev_arch.png
I use yEd to create the sketches.  It is nice b/c it is text based and fits well into repos.  In addition to getting things straight in my mind it helps to coordinate work with other and to make it more maintainable if folks have an idea of the high-level structure.  

I think you could start with the existing gazebo/gazebo_ros functionality to add a sensor to each joint, but it would be nice to automate the system such that a user could specify force feedback for an entire manipulator model and then the system would create the appropriate gazebo/gazebo_ros plugins for all the joints.  Might be able to do this with xacro or embedded ruby?

You might take a look at the HAPTIX project and ROS Industrial to make sure this doesn't exist somewhere.  Carlos wasn't aware of anything, but I'm surprised that it it isn't around and want to avoid reinventing any wheels.  If indeed isn't done it will be a really nice contribution.

Youssef-Khaky: These are good resources. Thank you.  
I will look into automating the entire manipulator arm's sensor attachment.

Youssef-Khaky: I think I am more in favour of the first one because it is applied to joints were the second one is applied to links. From my understanding, these kind of sensors are placed at the joints.  
Additionally, applying them at the joints means they have torque, the second one doesn't have torque since it is on links.  
I looked at couple of these sensors. The common configeration I have been noticing is putting those FT sensors at the base of an end-effector and not on every joint. Like [this](https://youtu.be/K5Z_pmBNrP8?t=65) video from Robotiq and [this](https://youtu.be/b4nz_hAh7qs?t=126) video from ATI.  Should we consider having it only at on the wrist joint?

bsb808: I agree.  Using the FT sensors seems like the way to go.

I like the Robotiq demo video - I think doing something similar in gazebo would be really useful.  Having a spring to push against is a good idea!

We want to have feedback from all the joints.  The existing hardware gives feedback to teleoperators for each joint and we anticipate that will be a useful part of u/w manipulation where visibility is very limited.

uuv_dave/launch/plot_manipulator.launch
```diff
@@ -0,0 +1,7 @@
+<!-- rqt_plot for force feedback wrist sensor -->
+<launch>
+    <node name="ft_sensors_plot" pkg="rqt_plot" type="rqt_plot" args="/rexrov/ft_sensor_wrist/wrench/force/x
```
bsb808, line 3: Nit - Can you clean up the line breaks and indentation.

uuv_dave/launch/plot_manipulator.launch
```diff
@@ -0,0 +1,7 @@
+<!-- rqt_plot for force feedback wrist sensor -->
```
bsb808, line 1: Start launch files with `<?xml version="1.0"?>` tag.

(Aside - need to check this on existing launch files in the repo.)

uuv_dave/launch/plot_manipulator.launch
```diff
@@ -0,0 +1,7 @@
...
+<launch>
+    <node name="ft_sensors_plot" pkg="rqt_plot" type="rqt_plot" args="/rexrov/ft_sensor_wrist/wrench/force/x 
+                /rexrov/ft_sensor_wrist/wrench/force/y
```
bsb808, line 4: I believe the torques maybe of more practical interest, since that is more typically measured.  I'll take a look at a demo to illustrate.

uuv_dave/worlds/uuv_dave_ocean_waves_spring.world
```diff
@@ -0,0 +1,210 @@
...
+      <pose>0 0 0 0 0 0</pose>
+    </include>
+
+    <!-- Bounding box with sea surface -->
+    <include>
+      <uri>model://ocean</uri>
+      <pose>0 0 0 0 0 0</pose>
+    </include>
+
+    <plugin name="underwater_current_plugin" filename="libuuv_underwater_current_ros_plugin.so">
```
bsb808, line 72: Can the current be removed to simplify this example?

uuv_dave/worlds/uuv_dave_ocean_waves_spring.world
```diff
@@ -0,0 +1,210 @@
...
+          <max>3.141592653589793238</max>
+          <mu>0.0</mu>
+          <noiseAmp>0.0</noiseAmp>
+        </vertical_angle>
+      </constant_current>
+    </plugin>
+
+    <plugin name="sc_interface" filename="libuuv_sc_ros_interface_plugin.so"/>
+
+    <model name="springy plate">
```
bsb808, line 104: Could we create this as a standalone model file and then include it in the world file?  This would make the springy plate model reusable.

uuv_dave/urdf/serial_arm.xacro
```diff
@@ -0,0 +1,287 @@
...
+          <hardwareInterface>EffortJointInterface</hardwareInterface>
+        </joint>
+        <actuator name="${joint}_motor">
+          <hardwareInterface>EffortJointInterface</hardwareInterface>
+          <mechanicalReduction>1</mechanicalReduction>
+        </actuator>
+      </transmission>
+    </xacro:macro>
+
+    <!-- Force Feedback Macro -->
```
bsb808, line 50: Let's put this macro in its own file.  Most of the rest of this file is specific to the oberon arm.  Having a standalone xacro will be more reusable.

uuv_dave/urdf/serial_arm.xacro
```diff
@@ -0,0 +1,287 @@
...
+    <xacro:macro name="joint_ft_sensor" params="joint_name namespace">
+      <!-- Enable the Joint Feedback -->
+      <gazebo reference="${namespace}/${joint_name}">
+        <provideFeedback>true</provideFeedback>
+      </gazebo>
+      <!-- The ft_sensor plugin -->
+      <gazebo>
```
bsb808, line 57: For completeness, might explicitly specify the other SDF parameters for the plugin.  It looks like the only missing one is "gaussianNoise":

See [gazebo_ros_ft_sensor.cpp](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_ft_sensor.cpp)

PR #17, opened 2020-05-22
Adding ROV example mesh

bsb808:
Should be able to run gazebo...
```
roslaunch gazebo_ros empty_world.launch
```
and then use the GUI to insert the `erov` model.  

Image of what you should see in Gazebo.
erov

And here is what the collision looks like.
erov_collision

The single box collision may cause issues when adding thruster, manipulators and tools, but we can address that later.


BruceMty:
Nice.  I'm wondering, would some square edges and a couple of bolts be more representative of a typical view?

phfatmonkey:
Hmm, I'm not sure why it smoothes the edges like that. I'll have to look at that.

bsb808:
Added Ryan's new mesh.

image

bsb808:
@BruceMty Would you have time to review this PR?

BruceMty:
This loads fine in Gazebo.  @bsb808 , do edges need to be right-angled for a cleaner look where arms and sensors are mounted?  Also, regarding my previous comment, I don't mind rounding, I'm wondering if it might be nice to add some "machine shop" look to it?

bsb808:
@BruceMty Yes, it is just a first cut.  We'll need to refine this and other vehicle models.  

Please approve when you are ready.

> This loads fine in Gazebo. @bsb808 , do edges need to be right-angled for a cleaner look where arms and sensors are mounted? Also, regarding my previous comment, I don't mind rounding, I'm wondering if it might be nice to add some "machine shop" look to it?

PR #19, opened 2020-05-27
Dvl dev

dtdavi1:
Updated the DVL model meshes to use the ones Ryan developed.  No changes to how anything works,  but if you run the demos from the Wiki, the Workhorse Navigator models will be using the Ryan-created ones for that DVL.  There are no launch-file demos of the others, but I've tested them all locally (the only change is the mesh in the visual element of the SDF or URDF file).


BruceMty:
Will this PR be withdrawn?  It looks like this moved to nps_uw_sensors_gazebo/models?

dtdavi1:
Pulling for now.  I'll resubmit after I delete all of the models that have been moved to the nps_uw_sensors_gazebo repository.

BruceMty:
Thanks!

PR #28, opened 2020-07-24
launching into velocity control mode by default

M1chaelM:
Teleoperating a UUV without velocity control mode enabled is probably only good for testing the stress level of the operator. This small PR causes our demo to use velocity control mode by default.

### DIFF
--- a/uuv_dave/launch/uuv_dave.launch
+++ b/uuv_dave/launch/uuv_dave.launch
@@ -6,6 +6,7 @@
     <arg name="namespace" default="rexrov"/>
     <arg name="set_timeout" default="false"/>
     <arg name="timeout" default="0.0"/>
+    <arg name="velocity_control" default="true"/>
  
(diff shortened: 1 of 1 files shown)


uuv_dave/launch/uuv_dave.launch
```diff
@@ -71,14 +71,21 @@ UUV launch for project DAVE
         <arg name="yaw" value="-1.8"/>
     </include>
 
-    <!-- joystick control for rexrov and oberon7 -->
+    <!-- Velocity teleop (UUV stays in position when joystick is not used) -->
```
BruceMty, line 74: Instead of commenting out, lets please add "if" and "unless" conditionals on an arg such as "suppress_stationary", ref. http://wiki.ros.org/roslaunch/XML

M1chaelM: OK, thanks. I added this.

PR #30, opened 2020-08-07
Glider hybrid whoi model

woensug-choi:
Testing pushing to a branch. Basic template done. Waiting for left side propeller dae and glider body dae without propellers.
Not ready to merge yet.


woensug-choi:
Will make new pull request with three glider models at once with branch name called 'glider_models'

PR #32, opened 2020-08-12
Initial bathy plugin migration from whoi

woensug-choi:
From WHOI's bitbucket repository I've extracted wind_bathy_plugin and renamed it as bathy_dave_plugin. The compile test passed. However, when launched using 'roslaunch bathy_dave_plugin bathy_dave.launch', it says 

> [Dbg] [bathy_plugin.cc:113] Loaded 0 bathymetry data sources.

The original repository requires their dropbox shared data something like 

> 2019catalyst_wind/data/bathymetry/ncei19_woodshole.epsg26987.grid.dae

which we don't have access to. I don't know how to obtain those data from NOAA (Which was the title of the trello card [Generating DEMs from bathymetric data](https://trello.com/c/IFttWWdo/149-generating-dems-from-bathymetric-data)). Maybe I missed those source codes in the repository?

From here, M.Jakuba's help would be appreciated about how to obtain bathymetry data sources.


woensug-choi:
Successfully generated bathymetry DB files with additional source codes received from Mike Jakuba. Lastest commit includes those model files in dave/models.

Current problems
1. According to the source.cc file, the bathy_plugin requires to be called not only as a standard plugin inside world file but also inside <model> definition for 'interval_s'
    - this makes error msg, But still works. Need to check with Mike about this.
2. current 'worldPos' does not change with initial spawn_model args.
    - Bathymetry is inserted into the world according to the current 'worldPos' and origin of Bathy DBs to calculate rows and columns for tiles. 
    - I believe this problem has little to do with bathy_plugin itself but due me being unfamiliar with l gazebo.


Insertion of bathymetry while running the simulator does not work. The 'worldPos' does not change with the model spawn positions. Help from anyone familiar with gazebo programming is appreciated. @M1chaelM Any suggestion to whom I should ask for help?

woensug-choi:
Ready to review,

Tutorial wiki at [Bathymetry Plugin Tutorial(https://github.com/Field-Robotics-Lab/dave/wiki/Bathymetry-Plugin)

Issue #25, opened 2020-06-11
Remove the dave/sonar_dave_plugins directory


BruceMty:
The dave/sonar_dave_plugins directory contains a copy of the uuv image_sonar_ros_plugin Sonar model.  It was to contain a custom Sonar model.  Now sonar models are managed elsewhere.  This Sonar model copy provides libimage_sonar_dave_plugin.so.  The only entry to this plugin is sonar_dave_uuv/urdf/sonar_snippets.xacro and its only entry is dave/uuv_dave/urdf/sensor_snippets.xacro.  Its significant use is to provide Sonar for the primary example:

    roslaunch uuv_dave uuv_dave.launch

The fix must include redirection to UUV's libimage_sonar_ros_plugin.so and to .xacro files so that this example still works.  We may wish to perform this fix when a new Sonar model becomes available.

bsb808:
Could we use the sonar plugin from uuv_simulator in the example (uuv_dave.launch) and then remove the copy of the image sonar pluign in this directory?

BruceMty:
@bbingham-nps @M1chaelM this issue may be closed when https://github.com/Field-Robotics-Lab/dave/pull/64 is merged.

bsb808:
@BruceMty Roger - please close the issue when you merge your PR.

PR #39, opened 2020-08-24
adding github action for continuous integration test

M1chaelM:
A github action to test pushed code, based on the one used for the [vrx repository](https://github.com/osrf/vrx/blob/master/.github/workflows/ci.yml).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,29 @@ jobs:
           repository: Field-Robotics-Lab/nps_uw_sensors_gazebo
           path: nps_uw_sensors_gazebo
 
+      - name: Checkout uuv simulator 
+        uses: actions/checkout@v2
+        with:
+          repository: uuvsimulator/uuv_simulator
+          path: uuv_simulator
+
+      - name: Checkout rexrov2
+        uses: actions/checkout@v2
+        with:
+          repository: uuvsimulator/rexrov2
+          path: rexrov2
+
+      - name: Checkout uuv manipulators 
+        uses: actions/checkout@v2
+        with:
+          repository: uuvsimulator/uuv_manipulators
+          path: uuv_manipulators
+
+      - name: Checkout eca a9
+        uses: actions/checkout@v2
+        with:
+          repository: uuvsimulator/eca_a9
+          path: eca_a9
 
       - name: Set up workspace
         run: |


M1chaelM:
@caguero would you mind taking a look? I am confused about the code_check.sh script. When you call it, it appears to have no arguments, so I thought it was just a placeholder. However, when I do the same thing in the repo it fails so I must be missing something.

M1chaelM:
Note: since we have not been using code_check.sh, one of our header files was not passing the tests it runs (gazebo_ros_image_sonar.hh). Instead of creating a separate PR for these minor adjustments, I've just included them here, and I'm adding @BruceMty to double check that I didn't mess this up (it still seems to build fine so I think it should be OK).

M1chaelM:
@caguero Sorry for all the chatter. It looks like I broke the testing while trying to figure out how to pull in the other repositories we need to build against. Do you know how to do this properly? If not I will come back and look into it later today.

BruceMty:
> Note: since we have not been using code_check.sh, one of our header files was not passing the tests it runs (gazebo_ros_image_sonar.hh). Instead of creating a separate PR for these minor adjustments, I've just included them here, and I'm adding @BruceMty to double check that I didn't mess this up (it still seems to build fine so I think it should be OK).

This file needs to go away, but we need to resolve issue https://github.com/Field-Robotics-Lab/dave/issues/25 first.  Changes to gazebo_ros_image_sonar.hh are fine if "roslaunch uuv_dave uuv_dave.launch" works.

M1chaelM:
> > Note: since we have not been using code_check.sh, one of our header files was not passing the tests it runs (gazebo_ros_image_sonar.hh). Instead of creating a separate PR for these minor adjustments, I've just included them here, and I'm adding @BruceMty to double check that I didn't mess this up (it still seems to build fine so I think it should be OK).
> 
> This file needs to go away, but we need to resolve issue #25 first. Changes to gazebo_ros_image_sonar.hh are fine if "roslaunch uuv_dave uuv_dave.launch" works.

Understood. Thanks Bruce.

M1chaelM:
@caguero Does this look OK with you? I'm hoping to get these tests in place before we merge the rest of the PRs leading up to release 2.0

M1chaelM:
Sorry @caguero but do you think you could look at this today? Thanks!

caguero:
> @caguero would you mind taking a look?
> 
> EDIT: Please disregard the question below. I found the paths in the script itself and I am no longer confused.
> 
> I am confused about the code_check.sh script. When you call it, it appears to have no arguments, so I thought it was just a placeholder. However, when I do the same thing in the repo it fails so I must be missing something.

That's the c++ static checker. It looks into some directories for C++ code.

caguero (approved):
Looks good to me and CI is :green_circle:

PR #41, opened 2020-08-29
Update install directories in CMakeLists

mabelzhang:
During a fresh installation of DAVE in a clean Docker container, two `CMakeLists.txt` files had to be modified, in order for model files to be found at compile time and runtime. I believe these changes should be universal, in order for DAVE to work out of the box.

To test this PR, do a clean clone of the repo into a new workspace and compile from scratch.

For `dave_manipulator`, the installation directories `meshes` and `robots` simply do not exist in the repository.
Before the change, I got compiler error:
```
CMake Error at cmake_install.cmake:56 (file):
  file INSTALL cannot find
  "/home/developer/dave_ws/src/dave/dave_manipulator/predator_description/meshes".
```

For `uuv_dave`, `urdf`, `models`, and `worlds` needed to be included in `install()`, otherwise some files were not found at run time.
Before the change, I got run-time errors:
```
No such file or directory: /home/developer/dave_ws/install/uuv_dave/share/uuv_dave/urdf/rexrov_oberon7.xacro
```
```
[FATAL] [1598070318.504041, 0.000000]: Error: specified file /home/developer/dave_ws/src/install/uuv_dave/share/uuv_dave/models/bar/model.sdf does not exist
```
```
[Err] [Server.cc:392] Could not open file[/home/developer/dave_ws/install/uuv_dave/share/uuv_dave/worlds/uuv_dave_ocean_waves.world]
```
All of these are about files not found in `install/`.

After the changes, things compiled and ran correctly for me, and I was able to see the robot on the seafloor when I ran
```
roslaunch uuv_dave uuv_dave.launch 
```

### DIFF
--- a/dave_manipulator/predator_description/CMakeLists.txt
+++ b/dave_manipulator/predator_description/CMakeLists.txt
@@ -11,6 +11,6 @@ catkin_package(
 #  DEPENDS system_lib
 )
 
-install(DIRECTORY meshes params robots urdf
+install(DIRECTORY params urdf
         DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
         PATTERN "*~" EXCLUDE)
--- a/uuv_dave/CMakeLists.txt
+++ b/uuv_dave/CMakeLists.txt
@@ -5,6 +5,6 @@ find_package(catkin REQUIRED COMPONENTS gazebo_ros)
 
 catkin_package()
 
-install(DIRECTORY launch
+install(DIRECTORY launch urdf models worlds
         DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
         PATTERN "*~" EXCLUDE)


M1chaelM:
@mabelzhang I can see that the `meshes` and `robots` directories don't exist and `urdf`, `models`, and `worlds` are not included in `install` but I don't get either the compiler error or the runtime errors you are seeing. Any idea why this might be the case?

mabelzhang:
Hmm that is interesting that it doesn't complain for you.

I have to admit I did not follow the wiki installation instructions closely, because I wanted to have Gazebo 9 from source, to be able to test development later. I had a Gazebo 9 workspace built using `colcon`, so I used `colcon` to build DAVE as well. There could be a difference there? Like `catkin`, `colcon` is cmake underneath. I don't know all the nuances of `colcon`, but it works for pure-cmake packages that aren't ROS packages.

Maybe different cmake versions (I have 3.10.2)? I was using Ubuntu 18.04, ROS Melodic, Gazebo 9.

The run-time things could be a difference between the two compilers. I think `catkin` might add the source directories to the search path, while `colcon` requires them to be installed.

Sorry, I guess it doesn't need to be changed if you don't intend for users to use `colcon`. Technically though, the install directories should correlate with what is desired to be installed.

M1chaelM:
Makes sense. I agree the install directories should correlate with what is desired to be installed and I don't see anything wrong with these changes. I just built against your branch to make sure. I'm approving so you can merge.

I do want to alert @Youssef-Khaky because I suspect he is about to add the meshes and robots directories back in his upcoming PR.

PR #42, opened 2020-08-30
Predator visuals

Youssef-Khaky:
In this PR I add the meshes obtained from Ryan to the robot arm. There are also few minor changes.

To test this PR, please run the following command.
```
roslaunch uuv_dave uuv_predator.launch
```

predator_arm

__Picture of Predator arm attached to UUV__

**Summary of changes:**
- Add the meshes of the Predator arm links. 
- Removed an extra joint from the control files (it was left over from the oberson7).
- Change some PID values and friction and damping to remove oscillations from gripper.

**Issues with the current arm:**  
1. No texture: The individual links were not centralized. After I centralized them in Blender, the tag in the dae file responsible for texture disappears.
2. Collision = visual. No individual collision models with reduced textures was produced. ∴ , which may result in more collision computations than needed.  
3. There is a slight tilt in the arm at the initial position of the azimuth angle. The initial azimuth angle is set to be 0, however that slight tilt means it is not exactly at 0.

### DIFF
--- a/uuv_dave/launch/uuv_predator.launch
+++ b/uuv_dave/launch/uuv_predator.launch
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <launch>
     <arg name="gui" default="true"/>
-    <arg name="paused" default="true"/>
+    <arg name="paused" default="false"/>
     <arg name="world_name" default="$(find uuv_dave)/worlds/uuv_dave_ocean_waves.world"/>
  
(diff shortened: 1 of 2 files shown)


bsb808:
I added a brief wiki to document at https://github.com/Field-Robotics-Lab/dave/wiki/Manipulator-Models

Also created issues #48 and #49 to address next steps.  For now, I can merge to keep this moving.

bsb808 (approved):
Works for me.

PR #43, opened 2020-08-30
Simple occlusion

dtdavi1:
Simple visual-based occlusion implementation example.  Use the following command to run the demo:

roslaunch uuv_dave simple_occlusion_demo.launch

What you'll see is similar to the grabbable bar tutorial, except when the bar falls out of nowhere, it will end up mostly burried in the "silt".  There is also a large cinderblock to the left of the UUV.  If you maneuver the UUV to the cinderblock and touch it (with the manipulator or just bump into it), you should see the UUV part penetrate the cinderblock's "moss" before colliding with the actual cinderblock.

The bottom occlusion (silt) is implemented by raising the visual mesh 0.15m above the collision mesh.  The cinderblock occlusion (soft moss) is implemented by increasing the scale of the visual mesh.

I still need to write up the tutorial, but probalby won't get it done before Wednesday's feature freeze, so I wanted to get the pull request in work sooner.


M1chaelM:
Great. I ran this and experienced the cinderblock moss and silt you describe.

PR #51, opened 2020-09-09
Static check tweaks

M1chaelM:
Very minor PR to update static code checker so it looks for .cpp extension, plus a small (probably irrelevant fix) to avoid checker complaint.

I don't think this requires a full test, but @caguero can you glance quickly at my change to code_check.sh and confirm that it makes sense to look for .cpp extensions as well as .cc, since we have both? Just want to make sure I'm not accidentally violating a convention I don't know about.


caguero (approved):
It looks reasonable to me!

M1chaelM:
Thanks @caguero for the quick review!

PR #57, opened 2020-09-11
adds .cpp files to cpplint check

M1chaelM:
One line change to fix issue where files with the .cpp extension were getting ignored by cpplint.py.

### DIFF
--- a/sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
+++ b/sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
@@ -31,24 +31,26 @@
  *
  */
 
-#include <algorithm>
 #include <assert.h>
+#include <sensor_msgs/point_cloud2_iterator.h>
+#include <sensor_msgs/image_encodings.h>
+#include <cv_bridge/cv_bridge.h>
+#include <tf/tf.h>
+
+#include <opencv2/core/core.hpp>
 #include <boost/thread/thread.hpp>
 #include <boost/bind.hpp>
 
-//#include <smarc_gazebo_ros_plugins/gazebo_ros_image_sonar.h>
+#include <limits>
+#include <string>
+#include <vector>
+#include <algorithm>
+
 #include <sonar_dave_uuv/gazebo_ros_image_sonar.hh>
 #include <gazebo/sensors/Sensor.hh>
 #include <sdf/sdf.hh>
 #include <gazebo/sensors/SensorTypes.hh>
 
-#include <sensor_msgs/point_cloud2_iterator.h>
-
-#include <tf/tf.h>
-
-#include <sensor_msgs/image_encodings.h>
-#include <cv_bridge/cv_bridge.h>
-#include <opencv2/core/core.hpp>
 
 namespace gazebo
 {
@@ -57,7 +59,8 @@ GZ_REGISTER_SENSOR_PLUGIN(GazeboRosImageSonar)
 
 ////////////////////////////////////////////////////////////////////////////////
 // Constructor
-GazeboRosImageSonar::GazeboRosImageSonar() : SensorPlugin(), width(0), height(0), depth(0)
+GazeboRosImageSonar::GazeboRosImageSonar() : SensorPlugin(), width(0),
+                                             height(0), depth(0)
 {
   this->point_cloud_connect_count_ = 0;
   this->depth_info_connect_count_ = 0;
@@ -115,8 +118,11 @@ void GazeboRosImageSonar::Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf)
   // Make sure the ROS node for Gazebo has already been initialized
   if (!ros::isInitialized())
   {
-    ROS_FATAL_STREAM_NAMED("depth_camera", "A ROS node for Gazebo has not been initialized, unable to load plugin. "
-      << "Load the Gazebo system plugin 'libgazebo_ros_api_plugin.so' in the gazebo_ros package)");
+    ROS_FATAL_STREAM_NAMED("depth_camera",
+                           "A ROS node for Gazebo has not been initialized, "
+                           "unable to load plugin. " << "Load the Gazebo "
+                           "system plugin 'libgazebo_ros_api_plugin.so' in "
+                           "the gazebo_ros package)");
     return;
   }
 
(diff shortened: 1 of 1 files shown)


M1chaelM:
Alas, once the check is enabled it finds 260 new errors that need to be resolved before it will pass it's own test. Working on this.

M1chaelM:
@BruceMty This is a minor change that required some stylistic tweaks to `gazebo_ros_image_sonar.cpp` to bring it in line with our checks. If you wouldn't mind looking over the changes to verify that nothing offends your sensibilities, that would be appreciated. If it looks Ok to you go ahead and approve.

sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
```diff
@@ -668,68 +703,74 @@ cv::Mat GazeboRosImageSonar::ComputeNormalImage(cv::Mat& depth)
...
+    cv::repeat(cv::Mat(t_x).reshape(1, 1), t_y.size(), 1, X);
+    cv::repeat(cv::Mat(t_y).reshape(1, 1).t(), 1, t_x.size(), Y);
     dist_matrix_ = cv::Mat::zeros(depth.rows, depth.cols, CV_32FC1);
     cv::multiply(X, X, X);
     cv::multiply(Y, Y, Y);
     cv::sqrt(X + Y + 1, dist_matrix_);
   }
 
   // TODO: make these into proper parameters
-  cv::Mat TS = intensity*images[2]; // target strength, probably dir should be DI
```
BruceMty, line 731: Recommend not removing comments that the writer put in.  It likely provides meaning and was placed intentionally.  Maybe move whole comment to line above.

sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
```diff
@@ -909,59 +966,70 @@ cv::Mat GazeboRosImageSonar::ConstructScanImage(cv::Mat& depth, cv::Mat& SNR)
...
   cv::Scalar blue(15, 48, 102);
   cv::Scalar black(0, 0, 0);
   cv::Mat scan(raw_scan.rows, raw_scan.cols, CV_8UC3);
   scan.setTo(blue);
-  
+
   cv::Point center(scan.cols/2, scan.rows);
   cv::Size axes(scan.rows+3, scan.rows+3);
-  cv::ellipse(scan, center, axes, -90, -fov/2., fov/2., black, -1); //, int lineType=LINE_8, 0);
+  cv::ellipse(scan, center, axes, -90, -fov/2., fov/2., black, -1);
```
BruceMty, line 978: Please leave in author's comment unless it is no longer applicable.  In this case, the comment suggests a modification to the action that is currently not chosen.

sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
```diff
@@ -909,59 +966,70 @@ cv::Mat GazeboRosImageSonar::ConstructScanImage(cv::Mat& depth, cv::Mat& SNR)
...
       }
-   }
+    }
   }
-  
+
   cv::Scalar white(255, 255, 255);
   cv::Size axes1(2./3.*scan.rows, 2./3.*scan.rows);
   cv::Size axes2(1./3.*scan.rows, 1./3.*scan.rows);
-  cv::ellipse(scan, center, axes, -90, -fov/2.-0.5, fov/2., white, 1, CV_AA); //, int lineType=LINE_8, 0);
```
BruceMty, line 945: Maybe put comment above this line.

sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
BruceMty: This file will be removed when issue 25 is addressed.

M1chaelM: Bruce, was this your comment? If so, could you clarify what it meant? There were many comments in the code that appeared to be "notes to self" and it seemed the file was in a draft state.

M1chaelM:
@BruceMty Just so you know, the way the github review functionality works is that if you make comments one at a time they are sent immediately to everyone who gets notifications. This is how Youssef accidentally spammed us. If you plan to make multiple comments and you want them to be released all at once, you can start a review, enter comments as part of the review, and then submit the whole thing when you are finished.

sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
BruceMty: No, all these comments were inherited.

But please disregard all this; I'll delete this file today as part of fixing issue 25.

sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
M1chaelM: (I believe the commented code would just set the parameters to their default values. You're right this is slightly more than a stylistic tweak.)

sonar_dave_plugins/sonar_dave_uuv/src/gazebo_ros_image_sonar.cpp
M1chaelM: I see. OK.

BruceMty:
I am concerned about the CI that we have put into place:

* It should not have been put in place in a way that breaks existing builds.  Merge requests are failing because of pre-existing CI failures relating to Lint checking.  This is blocking merges from happening and is stalling out our ability to go forward.

* Lint is enforcing policies that have no value or that are not standard practice in the Gazebo or ROS code base.  In particular, it enforces a maximum line length=80 policy.  For consistency, code should follow the coding style of the project for which it targets.

* Code we develop is adapted from other code.  Changing existing code to pass Lint does not harden existing code and we risk breaking things.  Also, changing the syntax of existing code makes it harder to spot semantic changes.

I recommend we tune Lint and the CI process to be helpful but not overly demanding.

M1chaelM:
@BruceMty Although I agree in spirit that the goal of the CI is to be helpful, I think these are pretty sensible and widely accepted style checks (especially length=80) and it's not too hard to fix the code to conform. I do wish we had put it in earlier, but since there are only a few files actually being checked, it doesn't seem like much of a burden to convert things over. Personally, I find the results much more readable.

That said, let's put this on our agenda for this week's meeting and see what others think.

M1chaelM:
@BruceMty Since this PR does not turn CI on or off, but rather just applies it uniformly to different file extensions (.cc and .cpp), can you approve? I have already made the necessary style changes, and you are planning to delete the file in question anyway. If we decide some files should not be checked we will need to modify the script so it accepts a whitelist. In any case, it shouldn't depend on whether the file extension is .cpp or .cc--this is too subtle and would very likely result in accidentally skipping files we intended to check.

BruceMty:
@M1chaelM good point, this PR is not contingent on CI switches or styling.

BruceMty (approved):

PR #62, opened 2020-09-17
New vehcile smilodon

dtdavi1:
Added EROV example vehicle.  I couldn't find CAD drawings or an operators manual, so the thruster locations are a bit of a guess (I put them in mesh positions that looked like possible thruster housings).  To run the example launch use

roslaunch caldus_gazebo/caldus.launch

You can control it with uuv_dave joy_thrusterop.py, but the namespace will need to be changed from "smilidon" to "caldus".  The thrusters should be mapped about the same as Smilidon, but my joystick doesn't match what everyone else is using, so can't verify that.


bsb808:
Good to go.  Thank you @dtdavi1 
I added an issue to make the joystick teleop more generalized https://github.com/Field-Robotics-Lab/dave/issues/63

We should get you a Logitech gamepad so we all have the same hardware.  We have many in the lab.

Merge when ready.

bsb808 (approved):

PR #76, opened 2020-10-06
Depth-dependent ocean current

woensug-choi:
The depth-dependent ocean current plugin is ready for PR review. 

### Notes
- Modified from `uuv_world_plugin` of `uuv_simulator`
- The original plugins are structured as `world plugin` defined at `.world` file. In order to obtain vehicle depth, new `model plugin` is added to be defined at each model `xacro` file.
- The database (table, csv file) is located at a directory alongside world files and its path configured at world file for `world plugin`
- The base_link name is required at `model plugin` which is defined at each model `xacro` file.

### Basic diagram for plugin structures

PluginDiagram

### Quickstart
**WARNING: fix csv database file path at world file**
- location : `uuv_dave/worlds/uuv_dave_ocean_waves_transientcurrent.world/<database>` 
```bash
# fix the csv database file path first !
roslaunch uuv_dave uuv_dave_transientcurrent_demo.launch
rostopic /hydrodynamics/current_velocity/twist/linear
```
if you are familiar with rviz configurations `/<model_name>/current_velocity_marker` will make an arrow indicating ocean current

### Configuration file locations
- `world plugin` configuration : 
   - `uuv_ws/src/dave/uuv_dave/worlds/uuv_dave_ocean_waves_transientcurrent.world`
```xacro
<plugin name="underwater_current_plugin" filename="libdave_underwater_current_ros_plugin.so">
      <namespace>hydrodynamics</namespace>
      <depth_dependent_current>
        <topic>current_velocity</topic>        <database>/home/eric/uuv_ws/src/dave/uuv_dave/worlds/transientOceanCurrentDatabase.csv</database> 
```
- `model plugin` configuration 
     - `uuv_ws/src/dave/uuv_dave/urdf/rexrov_oberon7.xacro`

```xacro
      <plugin name="dave_transient_current_plugin" filename="libdave_transient_current_plugin.so">
        <base_link_name>$(arg namespace)/base_link</base_link_name>
      </plugin>
```

- depth-dependent ocean current database 
   - `uuv_ws/src/dave/uuv_dave/worlds/transientOceanCurrentDatabase.csv`
   - 3 columns, North direction velocity [m/s], East direction velocity [rad], Depth [m]

## Issues to be discussed
- file path configuration for csv database  at `.world` file is not relative path. `world://` prefix did not work.

### DIFF
--- a/transient_current_dave_plugin/dave_world_plugins/CMakeLists.txt
+++ b/transient_current_dave_plugin/dave_world_plugins/CMakeLists.txt
@@ -52,7 +52,7 @@ include_directories(
 	include
   ${catkin_INCLUDE_DIRS}
 	${Boost_INCLUDE_DIRS}  
-  ${CMAKE_CURRENT_BINARY_DIR}/msgs
+  ${CMAKE_CURRENT_BINARY_DIR}
   ${GAZEBO_MSG_INCLUDE_DIRS}
 )
 


dtdavi1 (commented):
It seems to work as advertised based on the demo launch file (quite a current at the RexROV depth!).  Has the GaussMarkhovProcess plugin been changed from the UUV Simulator version?

Based on my reading of the code, the plugin still publishes a single current velocity, but at the depth of the vehicle.  As written, it won't allow for DVL current profiling (it won't break what's already there, but the solution will be based on the current at the vehicle depth rather than along the DVL beams), and I don't think that it will work correctly if there are multiple "underwater objects" in the world.  These may or may not need to be changed for the merge to be approved, but if they get merged in, we will want to write it up as an issue.

woensug-choi:
> It seems to work as advertised based on the demo launch file (quite a current at the RexROV depth!). Has the GaussMarkhovProcess plugin been changed from the UUV Simulator version?
> 
> Based on my reading of the code, the plugin still publishes a single current velocity, but at the depth of the vehicle. As written, it won't allow for DVL current profiling (it won't break what's already there, but the solution will be based on the current at the vehicle depth rather than along the DVL beams), and I don't think that it will work correctly if there are multiple "underwater objects" in the world. These may or may not need to be changed for the merge to be approved, but if they get merged in, we will want to write it up as an issue.

@dtdavi1 Thank you for the quick review! The Gauss-Markov process has not been changed at all. The transient current is added to the output of the Gauss-Markov process at every update. This means that the Gauss-Markov process is considered as background current of the database-derived ocean current (I believe it makes some sense considering statistical characteristics of the noise).

I think multiple issues can be raised from this initial PR including multiple vehicle supports and DVL support. Both issues are in relatively close reach but we need to discuss the priority. Any suggestions on how to organize the follow up extended features for later? Things that I come up to mind right now are as follows

### Future extended features
1. Match the database file format with common recording formats (no knowledge on the common format yet)
2. Multiple vehicles support
   - This feature was considered by having `model plugin` to give its `base_link` name to the world plugin. The gazebo topic under the curtain can be generated as many as the configurations at xacro model descriptions. Things to be added are (1) recognize multiple topics including certain prefix on air from world plugin. (2) Multiple publishers/subscribers from world plugin for each vehicle.
3. DVL sensor connect support
   - I am guessing that the DVL sensor topic is up on air. Having a bool tag to include a subscriber for the data won't affect the current PR.

transient_current_dave_plugin/dave_world_plugins/src/UnderwaterCurrentPlugin.cc
```diff
@@ -0,0 +1,358 @@
...
+{
+#if GAZEBO_MAJOR_VERSION >= 8
+  common::Time time = this->world->SimTime();
+#else
+  common::Time time = this->world->GetSimTime();
+#endif
+  // Calculate the flow velocity and the direction using the Gauss-Markov
+  // model
+
+  // Update current velocity
```
tdenewiler, line 284: How would you feel about making the current generation take place in a separate ROS node that has a `ServiceServer` interface. This would move the database loading and reading from this plugin into the ROS node.

The benefits of this approach would be to hide implementation details of how the current is calculated for a given location and depth. Other groups could then update that node, or write their own custom nodes that implement the same interface. This would allow for calculating the current based on physical models, or for looking up databases containing ocean currents in locations all around the world.

Some examples of ocean currents databases:

[World Ocean Circulation Experiment](https://www.nodc.noaa.gov/woce/)
[Global Ocean Currents Database](https://www.ncei.noaa.gov/access/data/global-ocean-currents-database/)
[Ocean Surface Current Analyses Real-time](https://www.esr.org/research/oscar/)

A possible service definition for the interface is

```
std_msgs/Time time  # Time of desired current, 0 is for most recent time
geographic_msgs/GeoPose geo_pose  # Pose where desired currents are located, typically a vehicle location
---
geometry_msgs/Twist twist  # Velocity of current
```

If you are interested in this I can work on providing a pull request.

woensug-choi: Thank you very much for the comments and links. The team is having continuous conversations over this topic. I believe that this will be dealt with sometime near future. Regrettably, the current PR is to provide initial basic functionality. We will keep track of the idea!

woensug-choi:
@dtdavi1 I've added a new ROS publisher with custom msg for database read from CSV database.
Now, the plugin has two separate ROS topics.
- `/hydrodynamics/current_velocity` 
   - A single 3 dimensional vector of velocities n vehicle body frame (1 body only)
   - Generic `geometry_msgs::TwistStamped` format. X, Y, Z velocities saved in twist.linear.x, twist.linear.y, twist.linear.z.
- `/hydrodynamics/stratified_current_velocity`
   - Arrays of 3 dimensional vectors and arrays of depths in NED world frame.
   - Custom `dave_world_ros_plugins_msgs::StratifiedCurrentVelocity` format. Consists of float32[] depths, velocity North, velocityEast, velocityDown

### Quick how-to
#### Comamnds
```bash
# fix the csv database file path first !
roslaunch uuv_dave uuv_dave_transientcurrent_demo.launch
rostopic /hydrodynamics/current_velocity/twist/linear
```
#### Results
```
depths: [0.0, 10.0, 20.0, 30.0, 40.0, 50.0, 60.0, 70.0, 80.0, 90.0, 100.0, 110.0]
velocityNorth: [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 2.0, 1.5, 1.0, 0.5, 0.5, 0.5]
velocityEast: [1.5, 2.0, 1.5, 1.0, 0.5, 0.0, 0.5, 1.0, 1.5, 1.0, 0.5, 2.0]
velocityDown: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
---
depths: [0.0, 10.0, 20.0, 30.0, 40.0, 50.0, 60.0, 70.0, 80.0, 90.0, 100.0, 110.0]
velocityNorth: [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 2.0, 1.5, 1.0, 0.5, 0.5, 0.5]
velocityEast: [1.5, 2.0, 1.5, 1.0, 0.5, 0.0, 0.5, 1.0, 1.5, 1.0, 0.5, 2.0]
velocityDown: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
---
# contd.
```

### Notes
- The plugin has another hidden publisher using a gazebo node with the same topic name (`/hydrodynamics/current_velocity`). I kept it there as referencing the plugin from the UUV simulator. Does the DVL plugin use the topic published from the gazebo node? I tried to make a custom msg for the database using `.proto` but did not finish it. Was not familiar with generating array message in `.proto` format. If you need this, it may take some time to finish.

woensug-choi:
@mabelzhang  Compiles fine on my machine. But cause error on Ubuntu CI compile checking with following... Could you give it a quick look?
```
/tmp/uuv_ws/src/dave/transient_current_dave_plugin/dave_world_plugins/include/dave_world_plugins/UnderwaterCurrentPlugin.hh:22:10: fatal error: StratifiedCurrentVelocity.pb.h: No such file or directory
 #include <StratifiedCurrentVelocity.pb.h>
```

transient_current_dave_plugin/dave_world_plugins/CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+	CATKIN_DEPENDS
+		gazebo_msgs
+		gazebo_dev
+)
+
+include_directories(
+	include
+  ${catkin_INCLUDE_DIRS}
+	${Boost_INCLUDE_DIRS}  
+  ${CMAKE_CURRENT_BINARY_DIR}/msgs
```
tdenewiler, line 55:
```suggestion
  ${CMAKE_CURRENT_BINARY_DIR}
```

tdenewiler:
I had build errors locally on Ubuntu 20.04/Noetic (I merged my noetic-devel branch into yours). I was able to fix the build error by changing the path to the include directories as suggested in a review comment. Hope that helps.

bsb808:
Pending changes to DVL plugin to fully test.

dtdavi1 (approved):
The updates work as advertised and can be safely merged.  There are a couple of things that I think should be changed, but they're not showstoppers--if the merge is completed, I'll write them up as issues.  Here's a quick summary:
1.  Figure out a better way to read the .csv file than using a hard-coded full path so that the world file doesn't have to be edited for it to work.  Moving it into the models directory might work as a stopgap.
2.  Remove the cross-dependency with the TransientCurrentPlugin model plugin.  This just amounts to changing the hydrodynamics/current_velocity functionality back to what was in the original UUV Simulator plugin.  This will provide some backwards compatibility and make everything work with multiple vehicles at different depths.
3.  Change the velocity fields in the StratifiedCurrentVelocity message from float32[] for velocityNorth, velocityEast, and velocityDown to a single Vector3[] field.  This is more in keeping with the rest of the codebase.

PR #78, opened 2020-11-16
Noetic devel

tdenewiler:


bsb808:
@tdenewiler We aren't ready to support/test Noetic yet.  Working with users to determine if this is a need - also other options for support.
@lauralindzey is using Noetic and might be able to test the support. 

@crvogt - Could you see if this PR works on our standard dev environment?

crvogt:
> @crvogt - Could you see if this PR works on our standard dev environment?

Will do!

tdenewiler:
@bsb808, I understand about not wanting to take on the maintenance burden of a new OS/ROS distribution. Are you open to merging these changes if there are no regressions on 18.04/Melodic?

bsb808:
@tdenewiler Yes.  Ideally we would have other community members using 20.04/Noetic that could test.  Would like to avoid breaking things for them as well.  I believe @lauralindzey and colleagues may be using such an env.

bsb808:
@tdenewiler Also, out of curiosity, what is the use-case for extending to Noetic?  We've had good success using docker so that you can support different environments for different projects.

tdenewiler:
The use case is to allow for running Dave natively on 20.04, to make it easier to port from ROS1 to ROS2, and to be prepared for end-of-life of 18.04. There are no awesome packages or libraries on Noetic that are not available on Melodic or anything like that. I can see that none of those use cases are critical in the short-term.

crvogt:
@bsb808 Tested on 18.04/Melodic and everything compiled fine.
Ran `roslaunch uuv_dave uuv_dave_electrical_mate.launch` and `roslaunch usbl_gazebo tutorial.launch` with no issues.

lauralindzey:
@bsb808 - These changes work for me. I prefer this method of determining when to use C++17 to the one I had come up with and gotten merged. 

I think that the changes required to run on Noetic should be possible to make in a backwards-compatible way (at least, I haven't found any that are needed and can't be), and would love to see them merged into master to prevent each group using 20.04 from having their own fork. I'm happy to help with reviewing changes if that helps.

bsb808:
@lauralindzey Fantastic.  If you don't mind to spot check the Noetic changes it would be perfect.  I invited you as a maintainer so that you can review/approve.

usbl_gazebo/CMakeLists.txt
```diff
@@ -1,17 +1,11 @@
...
-## Compile as C++11, supported in ROS Kinetic and newer
-# If running on focal, will be using Noetic + Gazebo11, which requires c++17
-execute_process(COMMAND lsb_release -cs
-  OUTPUT_VARIABLE RELEASE_CODENAME
-  OUTPUT_STRIP_TRAILING_WHITESPACE
-)
-if(RELEASE_CODENAME MATCHES "focal")
-  set(CMAKE_CXX_STANDARD 17)
-  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+if(NOT "${CMAKE_VERSION}" VERSION_LESS "3.16")
```
lauralindzey, line 4: I like determining c++17 vs c++11 by CMAKE_VERSION rather than the os, since it's more future proof. I guess it has the potential to change behavior for somebody on Melodic if they upgrade (but the 18.04 debs only have 3.10). I don't think this should be a problem, but I defer to @crvogt for that.

usbl_gazebo/CMakeLists.txt
```diff
@@ -1,17 +1,11 @@
...
+    set(CMAKE_CXX_STANDARD 17)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
 else()
-  add_compile_options(-std=c++11)
+    set(LOCAL_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
lauralindzey, line 8: Why was this change necessary? It compiled on my machine without it, and the auto-generated CMakeLists.txt seems to push people towards using `add_compile_options`.

lauralindzey (commented):
@bsb808 Sounds good! I'm happy to act as a supplemental reviewer for PRs that impact Noetic support, but I won't merge anything until/unless it already has an OK from the core team.

I can confirm that these changes compile on 20.04/Noetic and standard launch files run, but don't actually know of a launch file that tests the USBL model.

I'd like to better understand why @tdenewiler switched from using `add_compiler_options(-std=c++11` to using `set(LOCAL_CXX_FLAGS ...)` before recommending that it be merged.

usbl_gazebo/CMakeLists.txt
tdenewiler: Another benefit of using `CMAKE_VERSION` is that it works on more than just the standard Ubuntu distribution. It can also work on other Debian derivatives as well as Fedora.

usbl_gazebo/CMakeLists.txt
tdenewiler: I am open to reverting this change back to what it was. I was following a pattern we typically use for our own packages where we prefer to set compiler flags per target rather than globally. It's a pattern we have used for a long time, and I don't really know if it makes a difference to CMake anymore (10 years ago there were plenty of edge cases where it did make a difference...).

lauralindzey: Absent an actual problem, I'd be inclined to keep the same pattern used by the broader project.

crvogt: > I am open to reverting this change back to what it was. I was following a pattern we typically use for our own packages where we prefer to set compiler flags per target rather than globally. It's a pattern we have used for a long time, and I don't really know if it makes a difference to CMake anymore (10 years ago there were plenty of edge cases where it did make a difference...).

@tdenewiler I agree with @lauralindzey. If you could revert that line I'll do a review.

tdenewiler: I didn't realize you were waiting on that change. It's reverted in af2d636 and still builds and runs for me on Noetic.

crvogt (approved):
Tests:
- Compilation completes without issue
- `roslaunch usbl_gazebo tutorial.launch` runs as expected
- `roslaunch uuv_dave uuv_dave_electrical_mate.launch` runs as expected

PR #88, opened 2021-01-07
Actually install the bathy plugins

daewok:


woensug-choi (approved):
Tested. No error found.

Issue #85, opened 2021-01-06
Fix hard-coded path in stratified current demo launch file


dtdavi1:
In order to run the demo, the path to the .csv file specified in the launch file (uuv_dave/uuv_dave_transientcurrent_demo.launch) must be manually edited to a valid path on the local machine.  This needs to be changed to a relative path that the ROS or Gazebo "find" functionality can locate.

woensug-choi:
Jessica might work on this subject as a practice :)

j-herman:
I have this working locally.  If anyone needs it soon-ish, I can create a pull request, but otherwise I'll send it up when I have a working two-vehicle demo.

j-herman:
Ok, per recommendations/ comments I have updated the code.  I'm going to cancel the existing pull request since I'm now working in a branch on this repo.
The code will allow for three options:
1.  Specify the filename in the <database> tag.  The file must be located in the uuv_dave/worlds folder
2.  Specify the full path to the file in the <database> tag
3.  Comment out the database tag altogether, and the default /uuv_dave/worlds/transientOceanCurrentDatabase.csv will be loaded

j-herman:
Recommend closing based on approved PR #95

j-herman:
Closing per 2/8 discussion; feel free to reopen if needed

PR #93, opened 2021-02-02
fix issue #85 - generic file path for transient current DB

j-herman:
Fix for issue #85

### DIFF
--- a/transient_current_dave_plugin/dave_world_plugins/src/UnderwaterCurrentPlugin.cc
+++ b/transient_current_dave_plugin/dave_world_plugins/src/UnderwaterCurrentPlugin.cc
@@ -213,19 +213,19 @@ void UnderwaterCurrentPlugin::
   else
   {
     // To hardcode path:
-    // this->databaseFilePath = 
+    // this->databaseFilePath =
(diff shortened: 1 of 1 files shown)


woensug-choi:
@j-herman  Thank you for looking into this subject. I've tried to set-up a more generical way to designate the path myself before but without success. What you have done so far looks promising!

What we want eventually is that we would need a tag at .world file where the plugin is called up. Something like <current_database_csv>transientOceanCurrentDatabase</current_database_csv> that the user can change whatever CSV filename they want. Assuming they have multiple database CSV files and switch them if needed in the /uuv_dave/worlds folder. 

So, if we can make it convey the read filename and convey them in the UnderwaterCurrentPlugin.cc would be great.
Something like 
(1) Define string variable at UnderwaterCurrentPlugin.hh
private: std::string databaseFileName

(2) Read tag from SDF
At the Load function,
if (!_sdf->HasElement("current_database_csv"))
    this->databaseFileName = "transientOceanCurrentDatabase";
else
    this->databaseFileName =
        _sdf->GetElement("current_database_csv")->Get<std::string>();

(3) pass on to the path
this->databaseFilePath = ros::package::getPath(“uuv_dave”) + “/worlds/" + databaseFileName + ".csv";

Issue #77, opened 2020-10-13
GPS latitude and longitude seem to have their signs flipped


xydesa:
* `roslaunch uuv_dave uuv_dave.launch`
* `rostopic echo /rexrov/gps`
```bash
header: 
  seq: 641
  stamp: 
    secs: 21
    nsecs: 830000000
  frame_id: "rexrov/gps_link"
status: 
  status: 0
  service: 1
latitude: -56.7189407714
longitude: -3.51555967066
altitude: -93.0446776133
position_covariance: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
position_covariance_type: 3
---
```
* The signs on both latitude and longitude are flipped from what is defined in the world file: https://github.com/Field-Robotics-Lab/dave/blob/master/uuv_dave/worlds/uuv_dave_ocean_waves.world#L38
* They also seem wrong given the comment in the world file
  * [-56.7189407714, -3.51555967066](https://www.google.com/maps/place/56%C2%B043'08.2%22S+3%C2%B030'56.0%22W/@-40.6017017,0.1756212,3z/data=!4m5!3m4!1s0x0:0x0!8m2!3d-56.7189408!4d-3.5155597)
  * [56.7189407714, 3.51555967066](https://www.google.com/maps/place/56%C2%B043'08.2%22N+3%C2%B030'56.0%22E/@55.1902927,3.9973646,5.65z/data=!4m5!3m4!1s0x0:0x0!8m2!3d56.7189408!4d3.5155597)

This seems to be caused by this PR in the uuv_simulator repo: https://github.com/uuvsimulator/uuv_simulator/pull/250
In that PR, the maintainer stated that the long term fix was to replace the sensor plugin completely with one that does the necessary computation independent of Gazebo.

bsb808:
Thank you for the report.  Given our applications, we haven't paid much attention to the GPS publications.  It sounds like switching to the standard hector GPS plugin would be a better option for use-cases that would have satellite access.

We'll leave it on the list, but it will be a low priority.  Happy to review a PR from the community.

woensug-choi:
I think the opposite signs are correct as they are supposed to be. Since the origin moved from `0.0`, `0.0` to `56.71897669633431` and `3.515625`. The location of the vehicle is now at `-56.71897669633431`, `-3.515625`.

The issue on my hand is that the value is slightly off from what it's supposed to be.
- Expected value
[...]

PR #94, opened 2021-02-02
Cleanup of uuv_dave.launch example

bsb808:
The default example is `uuv_dave.launch`.  This PR cleans up a few loose ends to update that example.

* Removes image sonar (image_sonar_ros_plugin.so implemented in uuv_simulator) from the Rexrov `uuv_dave.launch` because the primary multibeam sonar sensor plugin is now here: https://github.com/Field-Robotics-Lab/nps_uw_multibeam_sonar 
* Removes the uuv_simulator GPS plugin from the example because 
    1. GPS doesn't work underwater and
    2. [Issue#77](https://github.com/Field-Robotics-Lab/dave/issues/77) and
    3. if you need a GPS, you should use a standard plugin like the hector gps plugin.

### DIFF
--- a/uuv_dave/launch/uuv_dave.launch
+++ b/uuv_dave/launch/uuv_dave.launch
@@ -97,9 +97,4 @@
           args="-file $(find uuv_dave)/models/bar/model.sdf -sdf -model grabbable_bar -z -90">
     </node>
 
-    <!-- rqt_image_view for image_raw_sonar -->
-    <node name="rqt_image_view_sonar" pkg="rqt_image_view"
-          type="rqt_image_view" args="/rexrov/depth/image_raw_sonar">
-    </node>
-
 </launch>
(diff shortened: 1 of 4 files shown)


woensug-choi:
@bsb808 Quick question.
I once thought of removing image_sonar_ros_plugin.so plugin in the `uuv_dave.launch`, but thought it would be something worthwhile for those who are not going to use multibeam sonar. I thought having multibeam sonar is going to be provided as an 'Optional' advanced feature. Should we still remove sonar entirely from our first-look `uuv_dave.launch`?

woensug-choi (approved):
`roslaunch uuv_dave uuv_dave.launch` with new `uuv_dave/urdf/rexrov_base.xacro`, which does not include GPU sensor works correctly as they are supposed to be.

PR #95, opened 2021-02-05
specify stratified current database filename from .world

j-herman:


woensug-choi (approved):
They work as they are supposed to be. And the description note in the world file is also appropriate. Great work! I've made changes on the wiki page about this.

Issue #86, opened 2021-01-06
Remove plugin cross-dependency in transient ocean current plugin.


dtdavi1:
The single ocean current value published by the transient current plugin (uuv_dave/UnderwaterCurrentROSPlugin.cc) is dependent on a separate model plugin to select the depth for which the depth is published.  Since the plugin has been updated to publish an array of currents for all of the depths, the single value no longer needs to be depth-specific.  Reversing this change will maintain reverse compatibility with models that assume a single homogeneous ocean current value.

woensug-choi:
@dtdavi1 I remember you saying that you are working on this. If you are not, or if you need help, please let me know!

dtdavi1:
@woensug-choi Yes, I've got this one for action.  It's actually done for the most part (it's not "multi-vehicle" ready, though), but I haven't pushed the commit yet.

Issue #84, opened 2021-01-06
Change stratified current message to use Vector3[] instead of separate float32[] for velocity values.


dtdavi1:
Using a Vector3 array instead of three float32 arrays is more in keeping with the de facto standards of the rest of the codebase.  The single velocity value generated by the same plugin is already published as a single Vector3 value.

woensug-choi:
@dtdavi1 I remember you saying that you are working on this. If you are not, or if you need help, please let me know!

dtdavi1:
@woensug-choi Yes, I've got it for action.  It's actually finished and committed in the branch we were working in, but I won't submit the merge request until the second issue is included.

woensug-choi:
Thank you very much!

PR #96, opened 2021-02-08
Depth dependent ocean current

dtdavi1:
Address issues [86](https://github.com/Field-Robotics-Lab/dave/issues/86) and [84](https://github.com/Field-Robotics-Lab/dave/issues/84).  Launch is via the uuv_dave_transientcurrent_demo.launch file.  Functionality will be the same as documented in [PR 76](https://github.com/Field-Robotics-Lab/dave/pull/76).


woensug-choi (approved):
They are working as supposed to be 👍 

At this point, the vehicle is only available to be affected by non-stratified single `/hydrodynamics/current_velocity`. Function to read `/hydrodynamics/stratified_current_velocity`, interpolate with current vehicle depth, and apply with the our version of `luuv_underwater_object_ros_plugin` considering multiple vehicle support is expected @j-herman

PR #110, opened 2021-06-11
Dockerfile for Noetic/Gaz11

bsb808:
This is a first draft of adding docker/rocker support for Noetic/Gazebo11.  It would be good to get some collaboration on this to converge on durable docker image that we can use to standardize the development environment.

Here is the wiki for setup and use: https://github.com/Field-Robotics-Lab/dave/wiki/noetic_docker

* @mabelzhang - Can you merge this with any more "standard" Noetic/Gaz11 image contents from OR? 
* @M1chaelM -  It would be nice if this was the same image we will use for VRX.  Once we get a prototype working maybe it is time to get serious and move all our Dockerfiles to a separate repo and integrate with Dockerhub?
* @crvogt - I wasn't able to successful test with Noetic (catkin_make fails with a Python2 error) - not sure if you are working on separate branches?
* @j-herman - Could you be our tester for the image and wiki to verify it works for you.


M1chaelM:
Tutorial looks very good from my end! I can launch smilodon.launch, so I think this is good to go. Some notes:

- It looks like rocker will replace our build and run bash scripts.
- The --home and --user commands are especially nice.
- Attempting to launch joy_thrusterop.launch failed with the following error:
```
[ERROR] [1623453330.221628794]: Couldn't open joystick /dev/input/js0. Will retry every second.
```
- A quick look at /dev/ shows the "input" directory was not mounted. Adding the rocker argument `--devices /dev/input/js0` resolves the issue. I've updated the tutorial to put this argument in.

M1chaelM (approved):
All looks good from my end, following minor tweak to the rocker argument in the tutorial. I will create an issue to work on aligning this image with our vrx image and setting up a dockerhub push action.

docker/DockerfileNoetic
```diff
@@ -0,0 +1,77 @@
+# Start with ROS base image
+FROM ros:noetic-ros-base
+
+# Tools useful during development.
+RUN apt update \
+ && apt install -y --no-install-recommends\
+        cppcheck \
```
mabelzhang, line 7:
```suggestion
        build-essential \
        cppcheck \
```
Also, we like to use alphanumeric ordering for long lists like this, to make items easy to find.

docker/DockerfileNoetic
```diff
@@ -0,0 +1,77 @@
...
+# setup environment
+ENV LANG C.UTF-8
+ENV LC_ALL C.UTF-8
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    qtbase5-dev \
+    libgles2-mesa-dev \
+    ${GAZ} \
+    lib${GAZ}-dev \
+    ros-${DIST}-xacro \
```
mabelzhang, line 53: These are some common tools. I don't know if they're necessary, if Michael is able to get everything working. I just see it in my standard Dockerfile.
Maybe you don't want these if you are looking to keep dependencies to a minimum. They can always be added later.
```suggestion
    python3-catkin-tools \
    python3-rosdep \
    python3-rosinstall \
    python3-rosinstall-generator \
    python3-wstool \
    ros-${DIST}-xacro \
```

docker/DockerfileNoetic
```diff
@@ -0,0 +1,77 @@
...
+# Gazebo
+RUN . /etc/os-release \
+    && echo "deb http://packages.osrfoundation.org/gazebo/$ID-stable `lsb_release -sc` main" > /etc/apt/sources.list.d/gazebo-latest.list
+
+
+# setup keys
+# ROS
+RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
+# Gazebo
+RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D2486D2DD83DB69272AFE98867170598AF249743
```
mabelzhang, line 42: I think you might be able to go without this key. Going by [this tutorial](http://gazebosim.org/tutorials?tut=install_ubuntu), you shouldn't need it. I don't have this key in my Dockerfile. But somebody else should test it.

mabelzhang (commented):
I added a few packages as requested, but if Michael is able to make it work without them, I'm not sure you strictly need these, especially if you want to keep the Dockerfile to a minimum.

A note on directory convention. We usually put one `Dockerfile` per directory, with the directory name being the name of the image, and the build and join scripts one directory above, e.g.
```
docker
|- join.bash
|- build.bash
|- run.bash
|- image1
   |- Dockerfile
|- image2
   |- Dockerfile
```
The images would be built from the `docker` directory with `./build.bash image1`.

The file name being `Dockerfile` also triggers some editors to automatically use syntax highlighting for Docker, which the current new file name doesn’t.

If you want to go fancy, some of us like to add a non-root user, like we do for the [SubT Dockerfiles](https://github.com/osrf/subt/blob/final_event/docker/subt_shell/Dockerfile), which takes a `user_id` from the [`build.bash` script](https://github.com/osrf/subt/blob/final_event/docker/build.bash). Some commands, like `rosdep update`, prefer non-root user. It gives you a home directory. Just a nice-to-have, up to your personal reference. Practically, it makes no difference, because the non-root user has `sudo`.

If you do want to combine with VRX, I'm seeing at least these additional dependencies in my personal Dockerfile I use for VRX + DAVE:
```
ros-noetic-robot-localization
ros-noetic-velodyne-simulator
```
and some others that might not be as necessary.

docker/DockerfileNoetic
woensug-choi: https://github.com/osrf/docker_images/issues/535#issuecomment-851505821

mabelzhang: Oh, I was referring to the `D2486...` one – I think that one might not be needed.
The expired key is the `C1CF...` one which has been extended.

j-herman:
I ran through the tutorial all the way from the beginning.  It eventually worked, and I was able to get the Smilodon launched.  Here's the Reader's Digest version of my test notes:

System updates

•	Baseline environment: 
   o	newly updated/upgraded Ubuntu 18.04/ROS Melodic/Gazebo 9
   o	older graphics card: GeForce GTX 480, driver version 390.143
   o	FRL repos already installed in uuv_ws; feature/bsb/docker_noetic branch for dave; all Carson’s noetic updates for uuv_simulator and uuv_manipulators included (it definitely doesn’t work without them – got that data point early!)

•	Installed docker per the tutorial link.  Set up via repos.  Success – version 20.10.7.

•	Installed Nvidia container toolkit per tutorial.  Had to downgrade to earlier versions and run a cuda 9.0 image to do the checkout.  Is this ok, or will there be compatibility issues later?  If not, we need to specify the no-older-than Nvidia driver version in system requirements.

•	Installed rocker per the tutorial link.  I just did the basic installation, but the instructions for installing from within a venv had me wondering: do we recommend/ require/ discourage operating within a python venv?  Might be useful to add a “don’t do this” recommendation to that step of the tutorial.

I successfully built the container following the instructions in the tutorial.

But then, when I tried to run things from inside the container, there were hiccups.

•	I had to clean out my .bashrc.  Conflicts with anything that sets up and runs a python venv, looks like, so some of my other workspace references were causing problems.  Sourcing melodic automatically also had to go.  

•	Figured out I needed to source noetic manually.  Can we add that to the container scripts somewhere, just for convenience?
  
•	Had to delete the existing build and devel folders from uuv_ws before catkin_make would run cleanly.  This is not a one-off – every time I switched between the container and the base system 18.04/melodic setup, I had to delete all the contents of uuv_ws except the src folder, in both directions, then make again from scratch.  This is only an issue if I want to keep going back and forth between versions within one workspace – checking backwards compatibility, that sort of thing.  Good troubleshooting tip, though.

Eventually, I did get the Smilodon launched.  No joy with the predator/ rexrov demos – still in work for the SDF collisions?

Trying to open a second terminal via the join.bash script does not work at the moment.  Error was:
`OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: “b3575f2ed9cd”: executable file not found in $PATH: unknown`
I think the issue is that the container just doesn’t have bash installed, so we have two options – either add it to the container, or modify the tutorial with:
Open additional terminals in the container via:
```
docker ps   % to confirm container name, if needed
docker exec -it <container name> /bin/bash
```
Assessment: docker image is good to go, with optional addition of bash.  Tutorial is usable for a marginally capable, highly motivated installer, but can be cleaned up to make it more generally accessible.

docker/DockerfileNoetic
```diff
@@ -0,0 +1,77 @@
+# Start with ROS base image
+FROM ros:noetic-ros-base
+
+# Tools useful during development.
+RUN apt update \
+ && apt install -y --no-install-recommends\
```
j-herman, line 6:
```suggestion
 && apt install -y --no-install-recommends\
              bash \
```

j-herman (approved):

docker/DockerfileNoetic
M1chaelM: @j-herman Thanks for all these notes. 

## host-image OS differences
I think you found a few things that are specific to users who are running a different OS on the host. This should work, but since you are using the --user and --home options of the rocker command you are mounting your local home directory which pulls in the configuration files for bionic/melodic/gaz9, and your workspace built for that OS. 
* Just curious: if you run `catkin_make clean` instead of manually deleting the build and devel folders, does that work?
* It seems like using the same .bashrc is not a good plan unless the host and image are very similar. We can add a bashrc specific to the image instead.
* To answer your question about the rocker installation, I would say a virtual environment is recommended but not required.

## `join.bash`
I verified that `bash` is installed on the container. To see this, run:
```
which bash
```
in the container terminal. It will show you bash it at `/usr/bin/bash`. This is also the reason your command `docker exec -it <container ID> /bin/bash` works. That command executes the file at `/bin/bash` (which is a symbol link to `/usr/bin/bash`) _inside_ the container. If `bash` were not installed in the container it would fail.

The problem you encountered is that the tutorial gives you the wrong argument for the `join.bash` script: `dave_noetic`. This is the name we gave the first image you built, but we forgot that `rocker` creates another image from this one and uses a default random hash for the name. This second image is the one that gets launched as a container (i.e. running version of an image).

The result is that so you are trying to join a container that isn't running at all (and therefore also not running `bash`). The quickest fix to this is to edit the tutorial to tell the user to tag the rocker image after it gets created. Longer term, we should probably fix up our `join.bash` script, because it is kind of hacky (it asks for the name of the image and tries to look up the name of the running container that was generated from that image, but this is not ideal because it is possible to run multiple containers from the same base image, and if someone were doing that the script would not be able to distinguish between them).

docker/DockerfileNoetic
M1chaelM: I think it's good to include them. We are not trying to make this image super lean.

M1chaelM:
> If you want to go fancy, some of us like to add a non-root user, like we do for the [SubT Dockerfiles](https://github.com/osrf/subt/blob/final_event/docker/subt_shell/Dockerfile), which takes a `user_id` from the [`build.bash` script](https://github.com/osrf/subt/blob/final_event/docker/build.bash). Some commands, like `rosdep update`, prefer non-root user. It gives you a home directory. Just a nice-to-have, up to your personal reference. Practically, it makes no difference, because the non-root user has `sudo`.

@mabelzhang We are adding a non-root user and home directory with rocker. Are you saying it would be also be helpful to have a non-root user during the build process, or just at runtime?

Also, I was hoping to either totally eliminate the `build.bash` script, or just use it to store a long command line call to `rocker`. What do you think? Do we still need it?

docker/DockerfileNoetic
j-herman: That makes much more sense!  I thought I had run which bash but going back through my full notes I can't find it.  Deleted my non-functioning suggestion and will revise my notes on the pull request.  catkin_make clean does not work correctly on my system (no idea why, been like that through many updates and reinstalls) but deleting build and devel does the same thing, right?

M1chaelM: `catkin_make clean` is not quite as thorough as `rm -rf build/ devel/`.  I thought it might be good enough, but apparently not. Thanks for checking.

mabelzhang:
I didn't realize you were using rocker until later. I should have edited my comment. So no, the user part is fine the way it is.

As for `build.bash`, I actually don't use `rocker` myself (I just use `build.bash`), so I don't have a recommendation there. A long command sounds helpful, if that eliminates the user having to copy-paste a long thing every time.

docker/DockerfileNoetic
bsb808: I reorganized to the recommend layout in bc9997dbfa - much clearer

docker/join.bash
```diff
@@ -1,11 +1,43 @@
 #!/usr/bin/env bash
 #
-# Typical usage: ./join.bash vrx
+# Typical usage: ./join.bash dave_nvidia
```
M1chaelM, line 3: Rocker currently does not have any way to specify the name of the image. It only allows you to give the name of the container, and it gives the image no name (other than the ID hash). If we want to continue to use the image name as the argument to `join.bash` the user will need to either look at the rocker output or use docker command-line tools to look up the correct image hash by checking the ID of the container.

There are a bunch of ways around this, but the easiest thing is probably to specify the container name using rocker and modify the join script to take a container name as an argument directly. Then it doesn't have to try to look up the container name from the image ID.

M1chaelM: For example, the rocker command would be:
```
rocker --devices /dev/input/js0 --dev-helpers --nvidia --x11 --user --home --name <CONTAINER NAME> --git dave_nvidia:latest
```
And the argument to the join script would be:
```
./join_bash <CONTAINER NAME>
```

docker/join.bash
```diff
@@ -1,11 +1,43 @@
...
 #
 
 IMG=$(basename $1)
```
M1chaelM, line 6: In the above approach this line could be replaced with:
```
CONTAINER_ID=$1
```
And the rest of the logic can be removed.

M1chaelM: @bsb808 Sure. It looks like you might have also accidentally added the emacs backup file `Docker~`. I'll remove it (unless I'm misreading the commit).

bsb808: Thank you @M1chaelM 
I'm going to go ahead and merge.

docker/DockerfileNoetic
M1chaelM: Just a single data point, but I was unable to get this to run without both keys.

Issue #111, opened 2021-06-11
Docker image organization and workflow


M1chaelM:
- It would be useful to have a baseline docker image that supports development for both vrx and dave. Right now these use different images.
- We may want to set up a separate repository for this image.
- A github action that builds and pushes this image to Dockerhub would be convenient.
- We may also want to ship our dave and vrx releases as part of prebuilt images, or make these available for use with cloudsim.
- This might warrant moving Dockerfiles out of the individual repos, depending on what kind of use cases or workflows are envisioned.
- To decide the above we also need to sketch out the workflows we are intending to support.

M1chaelM:
* Baseline images have been created
* A separate repository has been set up for them here: https://github.com/Field-Robotics-Lab/dockwater
* The github action requirement is captured in issue 1 of that repository: https://github.com/Field-Robotics-Lab/dockwater/issues/1
* We have decided not to ship releases prebuilt; for now the base images are sufficient.
* PR #132 removes docker from this repo.
When #132 is merged I will close this.

M1chaelM:
#132 is merged.

PR #114, opened 2021-07-20
Updating run.bash to be consistent with VRX workflow

bsb808:
Did small update the the associated docker wiki instructions and `run.bash` to be consistent with the same process for VRX.

Please test these instructions: https://github.com/Field-Robotics-Lab/dave/wiki/noetic_docker

### DIFF
--- a/docker/run.bash
+++ b/docker/run.bash
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 #
+# Copyright (C) 2018 Open Source Robotics Foundation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


mabelzhang:
While testing this, I noticed that `build.bash` also needs to be updated.

Currently, `build.bash` names the image `dave_nvidia` by default, checks a `--without-nvidia` arg, and renames to `dave` if necessary.
That applied to the Melodic Dockerfile, which is based on an NVIDIA Docker image.

For Noetic, we're not building on top of an NVIDIA image, we're injecting NVIDIA via Rocker.
So the Docker image name `dave_nvidia` and `--without-nvidia` option are now misleading for Noetic.

Regarding naming, it could just be called `$1`, which will be either `dave_melodic` or `dave_noetic`.
As for `build.bash`, I don't know if we want to update the Melodic image to also use Rocker so that the two images are consistent, or just have 2 different `build.bash` scripts, one for Melodic, one for Noetic.

M1chaelM (changes requested):
@mabelzhang and @bsb808 We now have several repositories that use the same test image, so I've moved the test images to their own repository: https://github.com/Field-Robotics-Lab/dockwater

This is following @bsb808 's suggestion from PR #110, and documented in issue #111. In the interest of keeping our installation environments in sync, I suggest we stop using the Docker-related code in this repository, including the build and run scripts, and switch over to the new common repository. This will require updated Docker install tutorials.

Once the Docker images are consistent, we should update our host-based install tutorials to produce comparable build environments. I've written this up in issue #116.

bsb808:
This is on hold pending @M1chaelM docker unification project.

M1chaelM:
> This is on hold pending @M1chaelM docker unification project.

I think we can now close this.

bsb808:
Closing PR b/c the new [dockwater](https://github.com/Field-Robotics-Lab/dockwater) project takes care of all of this!

PR #120, opened 2021-07-24
Add project subdirectories to include directories

mabelzhang:
This is for the reorg branch tested on Noetic.

I compiled with installation and fixed some problems in the paths.

In `dave_gazebo_ros_plugins`, there’s an install command for the directory `include/${PROJECT_NAME}`:
```
install(DIRECTORY include/${PROJECT_NAME}/
```
but that directory doesn’t actually exist. 

The `include` directories in all the `dave_gazebo_*` packages did not have subdirectories that are of the name of the package. I went ahead and added those. They are recommended for best practice. They make it clear in the `#include` line which package the headers came from, and the headers will get installed to the respective project directories in `install/`.


dtdavi1:
I haven't been able to get the Docker version running, so I'm not able to do the review.  As far as the missing directories go, I'd suggest commenting them out of the CMakeLists.txt files for now.  It's there so that the plugin libraries will build correctly, but some (most?) of the dave_gaebo_* packages don't have any plugins at the moment.

mabelzhang:
I think the subdirectories to `include/` are good to add, as good practice. Even if we don't add them now, we should add them later so that things can get installed in an organized directory structure.

bsb808:
This is a good addition to keep the organization consistent.  My testing was limited to compiling in the docker container and running an example: https://github.com/Field-Robotics-Lab/dave/wiki/vehicle_examples

bsb808 (approved):
Good to go.

PR #129, opened 2021-07-28
Fixing static check errors

M1chaelM:
Creating a draft so I can see the output of cppcheck on our current CI image.

PR #139, opened 2021-07-30
cpplint.py Header guard fix

woensug-choi:
`cpplint.py` has option [`--root`](https://github.com/Field-Robotics-Lab/dave/blob/d884368eda9dabd2d581b74471b0b6144f3eedb0/tools/cpplint.py#L108-L122) but it's hard to use them with `sh tools/code_check.sh`

I've modified the cpplin.py not to recognize parent path. (Mabel's comment](https://github.com/Field-Robotics-Lab/dave/pull/129#issuecomment-889064522)
https://github.com/Field-Robotics-Lab/dave/blob/efb191ebd978c7be826635141f23fde9c58327ac/tools/cpplint.py#L1554
- before : gazebo/dave_gazebo_ros_plugins/include/dave_gazebo_ros_plugins/ocean_current_plugin.h
- after : ocean_current_plugin.h


woensug-choi:
Wrong PR destination.

PR #140, opened 2021-07-30
cpplint.py header guard problem fix

woensug-choi:
`cpplint.py` has option [`--root`](https://github.com/Field-Robotics-Lab/dave/blob/d884368eda9dabd2d581b74471b0b6144f3eedb0/tools/cpplint.py#L108-L122) but it's hard to use them with `sh tools/code_check.sh`

I've modified the cpplin.py not to recognize parent path. (Mabel's comment](https://github.com/Field-Robotics-Lab/dave/pull/129#issuecomment-889064522)
https://github.com/Field-Robotics-Lab/dave/blob/efb191ebd978c7be826635141f23fde9c58327ac/tools/cpplint.py#L1554
- before : gazebo/dave_gazebo_ros_plugins/include/dave_gazebo_ros_plugins/ocean_current_plugin.h
- after : ocean_current_plugin.h

### DIFF
--- a/gazebo/dave_gazebo_model_plugins/src/ocean_current_model_plugin.cc
+++ b/gazebo/dave_gazebo_model_plugins/src/ocean_current_model_plugin.cc
@@ -319,9 +319,12 @@ void TransientCurrentPlugin::CalculateOceanCurrent()
     // find current depth index from database
     // (X: north-direction, Y: east-direction, Z: depth)
     int depthIndex = 0;
-    for (int i = 1; i < this->database.size(); i++) {
-      if (this->database[i].Z() > vehicleDepth) {
-        depthIndex = i; break;
+    for (int i = 1; i < this->database.size(); i++)
+    {
+      if (this->database[i].Z() > vehicleDepth)
+      {
+        depthIndex = i;
+        break;
       }
     }
 
@@ -336,7 +339,8 @@ void TransientCurrentPlugin::CalculateOceanCurrent()
     }
 
     // interpolate
-    if (depthIndex == 0) {  // Deeper than database use deepest value
+    if (depthIndex == 0)
+    {  // Deeper than database use deepest value
       northCurrent =
         this->database[this->database.size()-1].X();
       eastCurrent =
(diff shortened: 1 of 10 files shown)


M1chaelM:
@mabelzhang 's comment was that the short header style is used for header guards in ROS and Gazebo, so I would prefer to use the same cpplint.py file that they do unless we have a compelling reason to deviate. It could be that the one we are using is old, or that we aren't using it correctly. Here's what they are using for the Gazebo 11 repo: https://github.com/osrf/gazebo/blob/gazebo11/tools/cpplint.py

mabelzhang:
I'm not a power static check user either. We also follow different standards between ROS 2, Ignition, and Gazebo, so... if you want to just pick you favorite :woman_shrugging: 

Looks like this is the file we use for ROS 2 https://github.com/ament/ament_lint/blob/bfe815fbfb3f0c9b90ff70b95ea4b4b602f39398/ament_cpplint/ament_cpplint/cpplint.py#L1795-L1804
Guard for `rosbag2/rosbag2_cpp/include/rosbag2_cpp/reader.hpp` looks like `#ifndef ROSBAG2_CPP__READER_HPP_`.

Ignition uses this
https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/codecheck/cpplint.py
https://github.com/ignitionrobotics/ign-cmake/blob/ign-cmake2/codecheck/header_guard.rule
Guard for `ign-gazebo/src/systems/log/LogRecord.hh` looks like `#ifndef IGNITION_GAZEBO_SYSTEMS_LOGRECORD_HH_`

Gazebo 11 uses this
https://github.com/osrf/gazebo/blob/f54f95e7cad959d22d5af1549149da6e81c12bed/tools/cpplint.py#L1033-L1034
Guard for `gazebo/gazebo/sensors/GpuRaySensor.hh` looks like `#ifndef _GAZEBO_SENSORS_GPURAYSENSOR_HH_`

:woman_shrugging: My feeling is that the repo maintainers just decided what works best for them and went with it. So I don't think you need to follow anyone else's standard in particular. I'd just pick one that everyone on this project likes best.

M1chaelM:
>  My feeling is that the repo maintainers just decided what works best for them and went with it. So I don't think you need to follow anyone else's standard in particular. I'd just pick one that everyone on this project likes best.
* OK, for now I will copy Gazebo 11. It's easy and seems to fix the problem.
* For the good of humanity and our robot friends, I dream of a future where everyone at OSRF picks a single standard style that contributors can copy. :p

mabelzhang:
> I dream of a future
Me too...... Ignition and ROS 2 were the new starting points, and they (we? I wasn't there) failed.

M1chaelM (approved):
Switching to Gazebo 11 cppylint.py fixed this error and has the benefit of implementing the same style as at least some part of Open Robotics.

PR #160, opened 2021-10-29
Mesh distortion for object degradation

mabelzhang:
Blender Python script + tutorial for the mesh distortion part of the object degradation tool.

I include 2 different Coke can models from Ignition Fuel for comparison, because their triangle layout is very different, and that leads to different results after the distortion.
One of them (`Coke_Can`) I recentered and rescaled to real world size, so the file is significantly different from upstream.

The `fouling_rating` right now is just used as an arbitrary relative scale between 0 and 100. In this initial proof of concept, it isn't meant to bear resemblance to the actual appearance definitions.

## To test

Follow the README.
Since this is intended to be upstreamed soon, the tutorial is in the README instead of the Wiki, so that the README can be upstreamed as a tutorial in the `ign-common` repo. The Wiki will have a page to point to the README.

After you run the script, it should look something like this in Blender:

For `Coke`:

```
file_path = '/path/to/dave/models/dave_object_models/models/Coke/meshes/coke.obj'
object_prefix = 'LPCoke_Cube'
method = ['subdiv_mod', 'vert_rand', 'edge_subdiv']
fouling_rating = 20
import sys
sys.argv = ['distort.py', file_path, object_prefix, fouling_rating, method]
exec(open('/path/to/dave/urdf/scripts/mesh_distortion/distort.py').read());
```

2021-10-29-014156_3706x2049_scrot

If you switch to textured view (the globe button at the upper-right of the 3D view panel in the upper-left):

2021-10-29-014153_3706x2049_scrot

```
fouling_rating = 50
```
2021-10-29-015003_3706x2049_scrot_foulRating50

```
fouling_rating = 100
```
2021-10-29-015055_3706x2049_scrot_foulingRating100

For `Coke_Can`, I omit the subdivision modifier and only apply the latter two in `method`:

```
file_path = '/path/to/dave/models/dave_object_models/models/Coke_Can/meshes/coke_can.obj'
object_prefix = 'coke_can'
method = ['vert_rand', 'edge_subdiv']
fouling_rating = 20
import sys
sys.argv = ['distort.py', file_path, object_prefix, fouling_rating, method]
exec(open('/path/to/dave/urdf/scripts/mesh_distortion/distort.py').read());
```

2021-10-29-005129_3706x2049_scrot

If you switch to textured view (the globe button at the upper-right of the 3D view panel in the upper-left):
2021-10-29-005127_3706x2049_scrot

```
fouling_rating = 50
```
2021-10-29-015236_3706x2049_scrot_foulRating50

```
fouling_rating = 100
```
2021-10-29-015500_3706x2049_scrot_foulRating100

### DIFF
--- a/urdf/scripts/mesh_distortion/README.md
+++ b/urdf/scripts/mesh_distortion/README.md
@@ -64,7 +64,7 @@ object_prefix = 'coke_can'
 Set optional arguments. If not specified, the default will be used.
 ```
 fouling_rating = 20
-method = 'deform'
+method = ['subdiv_mod', 'vert_rand', 'edge_subdiv']
 ```
 
 Put the args into the input array:


j-herman (approved):
I ran the script in Blender 2.93.5 for both models at different levels of distortion and loaded one of the output models into Gazebo to verify end-to-end.  Looks good!
PSA for anyone else trying this - if you've updated your NVIDIA driver to 470.x, Blender may not run (OpenGL issue), so roll back to 460.x.

mabelzhang:
Thank you! You're a step ahead of me - I haven't generated the SDF yet :joy:

PR #163, opened 2021-11-03
ERB template for SDF with varying friction parameters

mabelzhang:
This is a small PR, though the line diff looks huge because of an OBJ mesh file.
It builds on #160 for biofouling.

This adds a basic ERB template to generate SDF files with different friction parameters, based on a parameter currently called `fouling_rating`. The SDF file wraps around the distorted mesh models generated by #160.

Since I'm not familiar with the maths of friction and don't know what the parameters mean, I've only put in a small proof of concept on `mu`. I'm open to feedback after the release as to how exactly we want to vary those parameters, what they mean, etc.

The main file is the ERB template `urdf/scripts/sdf_custom_properties/coke_template.erb`, with README in the same directory detailing usage.
Most other files are reference files (with `.ref` in the file name so that they don't get overwritten by generated files) as examples of generated output.

This is just an initial proof of concept, to keep PRs bite size, with future iterations expected.
I might have another PR before the release if I get custom XML tags done.

## To test

Follow the README at `urdf/scripts/sdf_custom_properties/README.md`.

You should see something like these in Gazebo for the 2 distorted coke cans:

2021-11-03-050410_3706x2049_scrot_Coke

2021-11-03-050501_3706x2049_scrot_Coke_Can

## A note on friction parameters

If people already have in mind (I don't) which exact friction parameters need to be varied, that would make it easy.

For those unfamiliar, including me, there are [friction_demo.world](https://github.com/osrf/gazebo/blob/gazebo11/worlds/friction_demo.world) and [torsional_friction_demo.world.erb](https://github.com/osrf/gazebo/blob/gazebo11/worlds/torsional_friction_demo.world.erb) examples upstream in Gazebo 11.

`<collision>` tag is documented here http://sdformat.org/spec?ver=1.8&elem=collision
I think the main tags of interest are `<friction>` (which has sub-tags of `<torsional>, <ode>, <bullet>` ) and maybe `<contact>` (which has sub-tags of `<ode>, <bullet>`, and a few other parameters).

One thing to note is that the `<ode>` tag is read by more than just ODE, but other engines as well unless other if-statements in the code indicate otherwise. This part is not very well documented. We have a ticket upstream saying the friction parameters in SDF need to be better documented (linked at the bottom of the README).

### DIFF
--- a/urdf/scripts/sdf_custom_properties/README.md
+++ b/urdf/scripts/sdf_custom_properties/README.md
@@ -54,3 +54,8 @@ roslaunch dave_demo_launch distorted_coke.launch
 ## Resources
 
 For parameters in the `<collision>` tag, see [`sdformat` specification](http://sdformat.org/spec?ver=1.8&elem=collision).
+
+In Gazebo 11, regardless of the physics engine, parameters in the `<ode>` tag
+are used for most engines.
+This is an implementation detail.
+More documentation upstream is required, as ticketed in [this issue](https://github.com/ignitionrobotics/sdformat/issues/31).


j-herman (approved):
Demo ran fine, with the can distorted as expected.  I did not regenerate the sdf from the erb, just confirmed the output file, but that should suffice for this simple example.

mabelzhang:
Thank you!

Issue #148, opened 2021-08-10
catkin_make vs. catkin build


mabelzhang:
Consider using `catkin build` for compiling packages.

`catkin build` has been a more modern version to `catkin_make` for a few years.

The only additional dependency is `sudo apt-get install python3-catkin-tools`.
The main difference is that `catkin build` builds each package in isolation, which is explained more in the first paragraph here https://catkin-tools.readthedocs.io/en/latest/migration.html

From a user experience point of view, `catkin build` produces much cleaner console output, and when you want to build one package, it really builds just that one package, instead of building a huge chain of other packages too. The former saves a lot of development time.

Typically, unless you are using some advanced build features, there isn’t much migration to be done, but the link above also has a migration guide.

mabelzhang:
Just a quick update. I've tried this in the dockwater Docker image and it compiles out of the box:
```
sudo apt-get update
sudo apt-get install python3-catkin-tools
catkin build
```

Give it a try.

After building once, try building a single package (maybe change something somewhere, not sure if that's needed):
```
catkin_make --pkg dave_object_models
```
vs
```
catkin build dave_object_models
```
You'll notice that the former still goes and builds many other packages, whereas the latter only builds that specified package and nothing else.

woensug-choi:
Tested on Host installation case. `catkin build` works without any modifications. It compiles fast since it generically utilizes all cores of the system to compile in parallel.

mabelzhang:
Thanks for trying it out!

To move forward on this ticket, it's just a matter of updating the installation tutorials, if people want to adopt `catkin build`.

woensug-choi:
I vote to update to `catkin build` ! It looks much better. If we are going to do this, we need to try to install everything again from the scratch on a clean machine to check everything is there. My log also has `pip3 install catkin_pkg`.

mabelzhang:
Created a PR in `dockwater` to address this https://github.com/Field-Robotics-Lab/dockwater/pull/11
That should give us a default Docker environment that can run `catkin build`.
I've tested this only with DAVE and not other repos that use dockwater.

M1chaelM:
@bsb808 I tested this on dockwater last night and left a note there, but just so it doesn't get overlooked: running
```
pip3 install osrf-pycommon
```
on a freshly built noetic container gives the error:
```
Requirement already satisfied: osrf-pycommon in /usr/lib/python3/dist-packages (1.0.0)
```
[...]

PR #172, opened 2021-12-14
Refactor dave_gazebo_world_plugins_msgs to its own package

mabelzhang:
Update: Oops. I compiled with the wrong hash of `ds_sim`. Still WIP.

Helps #148.

A protobuf message, `StratifiedCurrentVelocity.proto`, was and is built into a library `dave_gazebo_world_plugins_msgs`.
There's a new dependency of `ds_sim` on it. (Previously, it was only used in `dave_gazebo_world_plugins`, where the library lived.)
To get `catkin build` to work (#148), it was the easiest to refactor the protobuf message out to its own package, so that it can be used by an external package (`ds_sim`).

## To test

In the `dockwater` Docker environment, compile with
```
catkin build
```
This should compile everything.

For backward compatibility. this should still work as well:
```
catkin_make
```

## Note

That warning about
```
CMake Warning at /opt/ros/noetic/share/catkin/cmake/catkin_package.cmake:418 (message):
catkin_package() include dir
'/home/master/dave_ws/build/uuv_gazebo_plugins' should be placed in the
devel space instead of the build space
```
is still there.

It's talking about this line:
```
catkin_package(
INCLUDE_DIRS
  ${CMAKE_CURRENT_BINARY_DIR}
```
It doesn't like that the protobuf generated files are in `build` and wants them to be in `devel`. ROS messages generate message files in `devel`.

Short of specifying a custom path to protobuf to put the generated `.pb.*` files in `devel`, which is a path that can be specified by the catkin variable `CATKIN_DEVEL_PREFIX` (see [this page](http://docs.ros.org/en/jade/api/catkin/html/user_guide/variables.html)), protobuf will generate files in `build`. I didn't see a way to tell protobuf a custom destination to generate files. `PROTO_SRCS` and `PROTO_HDRS` seem to be return values rather than input parameters to `PROTOBUF_GENERATE_CPP`.

Going by [this documentation](https://cmake.org/cmake/help/latest/module/FindProtobuf.html), it seems customary to put protobuf files in `${CMAKE_CURRENT_BINARY_DIR}`.

So I left the warning there.


mabelzhang:
Actually, the fix is much simpler, just a few lines in `ds_sim`, no refactoring needed.
Superseded by https://github.com/Field-Robotics-Lab/ds_sim/pull/5

The note above about the devel space warning is still true for the original package.

There were a few unrelated fixes in this PR that I still recommend - mainly removing `/usr/local/lib/` from the `CMakeLists.txt` file. That line is hardcoded and isn't needed for the package to compile. There were also some `\t` tab characters that should be spaces.

Issue #147, opened 2021-08-10
Use <author> tag in package.xml


mabelzhang:
During the recent repo reorganization for release 4.0.0, I noticed that the <author> tag is not used in `package.xml` files. Something to consider. Even though `git blame` shows the author per-line, commit histories for files are [not preserved when files are moved](https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history) (run `git log` on moved files, and you'll see all the history is lost).

For open source projects, arguably, it's important to credit the original authors. We don't do that in Gazebo / Ignition as there's no package.xml, but in ROS, we often have multiple <author> and multiple <maintainer> tags per package, for example, https://github.com/ros/robot_state_publisher/blob/noetic-devel/package.xml .

(We even automatically append authors of PRs for each release to the CHANGELOGs ([example](https://github.com/ros/robot_state_publisher/blob/noetic-devel/CHANGELOG.rst)) via Bloom, but Bloom is for releasing packages into rosdistro and not applicable here.)

Since the DAVE repo seems to have different contributors with their own expertise, and some have left the project, this is something to think about.

bsb808:
Proposed authors: https://docs.google.com/spreadsheets/d/1RsEDT80HVvjFNxyU4ATySshFTjZSPKICIs6GfQ18HQA/edit?usp=sharing

bsb808:
Addressed with PR #173

PR #173, opened 2021-12-15
Adding <author> tags to package.xml files

bsb808:
Addresses Issue #147.

The <author> tags are not required, but nice to have.

To test, just re-make and run your favorite dave example.

### DIFF
--- a/models/dave_object_models/package.xml
+++ b/models/dave_object_models/package.xml
@@ -3,7 +3,6 @@
   <name>dave_object_models</name>
   <version>3.1.1</version>
   <description>The dave_object_models package</description>
-  <author email="mabel@openrobotics.org">Mabel Zhang</author>
   <maintainer email="dtdavi1@nps.edu">dtdavis</maintainer>
   <license>Apache 2.0</license>
 
--- a/models/dave_worlds/package.xml
+++ b/models/dave_worlds/package.xml
@@ -3,7 +3,7 @@
   <name>dave_worlds</name>
   <version>3.1.1</version>
   <description>The dave_worlds package</description>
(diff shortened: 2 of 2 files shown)


models/dave_worlds/package.xml
```diff
@@ -3,7 +3,7 @@
   <name>dave_worlds</name>
   <version>3.1.1</version>
   <description>The dave_worlds package</description>
-
+  <author email="mabel@openrobotics.org">Mabel Zhang</author>
```
mabelzhang, line 6: I can't claim authorship here. I only added 1 file to the package very recently - probably why last modified shows me.

models/dave_object_models/package.xml
```diff
@@ -3,7 +3,7 @@
   <name>dave_object_models</name>
   <version>3.1.1</version>
   <description>The dave_object_models package</description>
-
+  <author email="mabel@openrobotics.org">Mabel Zhang</author>
```
mabelzhang, line 6: I can't claim authorship here. I only added 1 file to the package very recently - probably why last modified shows me.

models/dave_worlds/package.xml
bsb808: Switched to @dtdavi1 with 7658a50

models/dave_object_models/package.xml
bsb808: Removed author tag with 7658a50

dtdavi1 (approved):
Everything looks reasonable.  We might consider including multiple author tags in some of the packages (most of the ones that list me as the author actually have more content from other people than from me), but people should be able to add themselves as they change other things easily enough going forward..

PR #178, opened 2022-01-13
Beautiful bathy tiles with ogre heightmap but cannot change the tile

woensug-choi:
image

# This branch includes
- Pipelining the much beautiful bathymetry tiles rendered with ogre's heightmap feature
- Sample 4 bathymetry files
  - Generated using the Bathymetry_Converter. Textures are also automatically generated when dissecting into tiles
- Include a rospy node/launch that initiate the position of the vehicle using lat/lon
- A launch/world setting that will drift the vehicle to the side edge to cause the bathymetry plugin to trigger tile change

# To test
```
sudo apt-get install python3-gdal
roslaunch dave_demo_launch dave_bathymetry_demo.launch
```

# Problem
- The signal from the bathymetry plugin seems to be triggering the bathymetry tile ok. But nothing happens.
- Removing the heightmap on the Gazebo's left panel does not remove the heightmap
- It seems a scene can only have one heightmap (https://github.com/osrf/gazebo/issues/481).

@mabelzhang Is there any trick to remove and insert heightmap during the run? Is there anyone I can ask about this?


woensug-choi:
Found a method to make fairly beautiful bathymetry tile with OBJ mesh files too. Now updated at https://github.com/Field-Robotics-Lab/Bathymetry_Converter. Closing this PR which was attempt to utilize heightmap feature of the ogre.

PR #183, opened 2022-01-21
Fix typos and install models

mabelzhang:
- Cleans up small typos
- Installs model directories
- Remove unnecessary lines in dave_demo_launch CMake, since that package doesn't compile anything


j-herman (approved):
Compiled and ran the basic demo - all looks good!

mabelzhang:
Thanks for the review!

PR #185, opened 2022-01-29
Update object degradation tutorial; rename fouling_rating to distort_extent

mabelzhang:
I tagged @j-herman (because you've run this in the past) or @woensug-choi (because you're looking into this now) for review. Only one is needed.

- Update object degradation tutorial to include 1. how to add more degradation methods to the Blender Python script, and 2. how to add more custom SDF tags.
  The current SDF tags are just dummy examples. @woensug-choi if you find more suitable terminology, feel free to update the example - remember to also update the Gazebo C++ plugin, see README section "To add new custom tags". Alternatively, just use whatever you like in the multibeam sonar plugin and the SDF it loads.
- Rename `fouling_rating` variable in range [0, 100] to `distort_extent`, now in range [0, 1]. This follows the discussion that we don't care about actually adhering to the navy fouling scale. It's now just an arbitrary relative scale.

## To test
Follow the mesh README to run the mesh modification script in Blender 2.92 and make sure it still works after the variable name change.

### DIFF
--- a/urdf/scripts/mesh_distortion/README.md
+++ b/urdf/scripts/mesh_distortion/README.md
@@ -5,10 +5,13 @@ the generalization of machine learning algorithms.
 
 ## Prerequisites
 
-These instructions are written for [Blender](https://www.blender.org/) 2.92,
-which has a revamped user interface.
-The version may be newer than the version in `apt-get`.
+These instructions have been tested in [Blender](https://www.blender.org/) 2.92,
+which has a revamped user interface from previous versions.
+The version in `apt-get` may be older.
+These instructions should work in newer versions.
 Install manually as needed.
+If newer versions do not work for you, Blender 2.92 can be found
+[here](https://download.blender.org/release/Blender2.92).
 
 Helpful tip for Blender Python development:
 To show Python API in the tooltips when the cursor is hovered over a button or
@@ -35,6 +38,9 @@ Choose Transform while importing: Z Up, X Forward.
 
 Tip: To view the textured object in the viewport, click on the globe icon in the
 upper right of the viewport, which is for Viewport Shading: Material Preview.
+
+If the texture is not loading correctly for you, check that you have the entire
+folder for the object, which includes the `meshes` and `materials` directories.
 
 Look for the mesh name in the model.
 You can find this in the View Layer list in the upper-right panel.
--- a/gazebo/dave_gazebo_model_plugins/include/dave_gazebo_model_plugins/custom_surface_properties.h
+++ b/gazebo/dave_gazebo_model_plugins/include/dave_gazebo_model_plugins/custom_surface_properties.h
@@ -46,8 +46,8 @@ namespace gazebo
     /// \brief Surface roughness, [0.0, 1.0]
     private: double roughness = 0.0;
 
-    /// \brief Biofouling rating, [0, 100]
-    private: int biofoulingRating = 0;
+    /// \brief Distort extent, [0, 1]
+    private: float distortExtent = 0.0;
   };
 }
 
--- a/gazebo/dave_gazebo_model_plugins/src/custom_surface_properties.cc
+++ b/gazebo/dave_gazebo_model_plugins/src/custom_surface_properties.cc
@@ -37,13 +37,13 @@ void CustomSurfaceProperties::Load(physics::ModelPtr _parent,
   sdf::ElementPtr modelElt = _sdf->GetParent();
   this->material = modelElt->Get<std::string>("surface_props:material");
 
-  this->biofoulingRating = modelElt->Get<int>(
-    "surface_props:biofouling_rating");
+  this->distortExtent = modelElt->Get<int>(
+    "surface_props:distort_extent");
 
   this->roughness = modelElt->Get<double>("surface_props:roughness");
 
   gzdbg << "Read custom SDF tags: material " << this->material
-    << ", biofouling rating " << this->biofoulingRating
+    << ", distort extent " << this->distortExtent
     << ", roughness " << this->roughness << std::endl;
 }
 
--- a/urdf/scripts/mesh_distortion/distort.py
+++ b/urdf/scripts/mesh_distortion/distort.py
@@ -204,12 +204,13 @@ def distort(file_path, object_prefix, distort_extent, method):
     # Export result to file
     out_path = os.path.splitext(file_path)[0] + '_distort' + \
         os.path.splitext(file_path)[1]
-    # TODO: COLLADA is not exporting texture correctly, not sure why
+    # TODO: COLLADA is not exporting texture correctly, not sure why.
+    # TODO: collada_export() does not expose relative path option.
     if out_path.lower().endswith('dae'):
         bpy.ops.wm.collada_export(filepath=out_path)
     elif out_path.lower().endswith('obj'):
-        bpy.ops.export_scene.obj(filepath=out_path, axis_forward='X',
-            axis_up='Z')
+        bpy.ops.export_scene.obj(filepath=out_path, path_mode='RELATIVE',
+            axis_forward='X', axis_up='Z')
     else:
         print('ERROR: Only COLLADA (.dae) and OBJ formats are supported for exporting at the moment.')
         return
--- a/urdf/scripts/sdf_custom_properties/README.md
+++ b/urdf/scripts/sdf_custom_properties/README.md
@@ -55,14 +55,13 @@ on custom SDF elements and attributes can be found on the sdformat website.
 Currently, these custom properties are defined:
 
 - `<surface_props:material>`: `string`
-- `<surface_props:biofouling_rating>`: `int`, in range `[0, 100]`, defined by
-  the Navy biofouling scale.
+- `<surface_props:distort_extent>`: `float`, in range `[0, 1]`.
 - `<surface_props:roughness>`: `double`, in range `[0.0, 1.0]`
 
 Example SDF snippet:
 ```
 <surface_props:material>metal</surface_props:material>
-<surface_props:biofouling_rating>30</surface_props:biofouling_rating>
+<surface_props:distort_extent>30</surface_props:distort_extent>
 <surface_props:roughness>0.3</surface_props:roughness>
 ```
 
@@ -103,5 +102,3 @@ More documentation upstream is required, as ticketed in [this issue](https://git
 
 - Write a regression test to diff between the ERB generated file and the
   reference SDF file in the repo.
-- Flesh out the set of friction parameters that the biofouling rating should
-  affect
--- a/urdf/scripts/sdf_custom_properties/coke.ref.sdf
+++ b/urdf/scripts/sdf_custom_properties/coke.ref.sdf
@@ -83,7 +83,7 @@
 
       <!-- Custom SDF elements for surface properties -->
       <surface_props:material>metal</surface_props:material>
-      <surface_props:biofouling_rating>10</surface_props:biofouling_rating>
+      <surface_props:distort_extent>10</surface_props:distort_extent>
       <surface_props:roughness>0.1</surface_props:roughness>
       <plugin name="custom_surface_properties"
         filename="libcustom_surface_properties.so"/>
--- a/urdf/scripts/sdf_custom_properties/coke_template.erb
+++ b/urdf/scripts/sdf_custom_properties/coke_template.erb
@@ -17,9 +17,9 @@
       <uri>model://sun</uri>
     </include>
 
-<%# Example of how fouling_rating can be used to vary coefficients %>
+<%# Example of how distort_extent can be used to vary coefficients %>
 <%
-      fouling_rating = 10
+      distort_extent = 10
       mu_factor = 0.001
 
       n_objs = 1
@@ -61,7 +61,7 @@
                 </ode>
               </torsional>
               <ode>
-                <mu><%= 1.0 + fouling_rating * mu_factor * idx %></mu>
+                <mu><%= 1.0 + distort_extent * mu_factor * idx %></mu>
                 <mu2>1.0</mu2>
                 <fdir1>0 0 0</fdir1>
                 <slip1>0.0</slip1>
@@ -89,8 +89,8 @@
 
       <!-- Custom SDF elements for surface properties -->
       <surface_props:material>metal</surface_props:material>
-      <surface_props:biofouling_rating><%= fouling_rating %></surface_props:biofouling_rating>
-      <surface_props:roughness><%= fouling_rating * 0.01 %></surface_props:roughness>
+      <surface_props:distort_extent><%= distort_extent %></surface_props:distort_extent>
+      <surface_props:roughness><%= distort_extent * 0.01 %></surface_props:roughness>
       <plugin name="custom_surface_properties"
         filename="libcustom_surface_properties.so"/>
     </model>


woensug-choi:
Is it correct that the current subdivision level from 0~4 is the number of operations to divide an edge into two with a mid vertex?

Are there methods we can define the density of the random vertexes generated by subdivision?  Something like max edge lengths? (Still discussing what we need. No need to include in this PR :))

mabelzhang:
> divide an edge into two with a mid vertex?

I played with it a bit on a cylinder and a cube, and it does look like what you described.

Blender's official [description](https://docs.blender.org/manual/en/latest/modeling/modifiers/generate/subdivision_surface.html) doesn't say too much about the mathematical operation. "The Subdivision Surface modifier (often shorten to “Subdiv”) is used to split the faces of a mesh into smaller faces, giving it a smooth appearance." For the "Simple" mode, which is the current mode in the script, it "Only subdivides the surfaces, without any smoothing."

For a cylinder, the top/bottom cap looks like this:

level 0
2022-01-31-185155_3706x2049_scrot_cylinderSubdiv0

level 1
2022-01-31-185150_3706x2049_scrot_cubeSubdiv1

level 2
2022-01-31-185146_3706x2049_scrot_cylinderSubdiv2
2022-01-31-184413_3706x2049_scrot_cylinderSubdiv2

mabelzhang:
> density of the random vertexes generated by subdivision? Something like max edge lengths?

Hmm there might be a way to script that, although I don't know if edge length can be a parameter to easily control.
The Subdiv modifier doesn't seem to expose such a parameter. There might be a different way to do it.

Reference: [Subdiv modifier options](https://docs.blender.org/api/current/bpy.types.SubsurfModifier.html)
The "Simple" mode Blender says is equivalent to the mesh operation Subdivide, which has these options https://docs.blender.org/api/current/bpy.ops.mesh.html#bpy.ops.mesh.subdivide

mabelzhang:
You didn't ask about this, but I guess you've already seen it and that's why you mentioned bump height:
Vertex randomization has a parameter `offset` for "Distance to offset"
https://docs.blender.org/api/current/bpy.ops.transform.html#bpy.ops.transform.vertex_random

```
bpy.ops.transform.vertex_random(offset=0.0, uniform=0.0, normal=0.0, seed=0, wait_for_input=True)
Randomize vertices

Parameters
offset (float in [-inf, inf], (optional)) – Amount, Distance to offset

uniform (float in [0, 1], (optional)) – Uniform, Increase for uniform offset distance

normal (float in [0, 1], (optional)) – Normal, Align offset direction to normals

seed (int in [0, 10000], (optional)) – Random Seed, Seed for the random number generator

wait_for_input (boolean, (optional)) – Wait for Input
```

woensug-choi:
Yes, the script you've made can designate offset (=bump height) magnitude (while they are also random). but cannot control the density. 

I was just asking for the density to calculate statistics for the minimum mesh edge length which would lead to average normal angles. Hmm

mabelzhang:
The edge length could vary a lot depending on where you are on the mesh. With a cylinder, the edges of the circles are very small, while the edges on the wall can be very long for e.g. a pipe.

Do you just want the normals? Blender has other ways to get the normals, e.g. https://docs.blender.org/api/current/bpy.types.Mesh.html

woensug-choi:
Don't understand how the screenshot you've shown is possible. Maybe because it went through the process `subdivide`->`offset`->`subdivide`->`offset` order? What if finish `subdivide` and `offset`? 

What I would like to have is some statistics on how much the mean normals have deviated from the original object shape. I was imagining the `vertex density` that are produced out of subdivision could provide information about `mean normal deviation`. 

Or maybe we could ditch `subdivision` function called to whole object and use a manual loop for each mesh elements and give custom subdivision density.

mabelzhang:
To clarify, the screenshot I showed in the meeting came straight from a single operation of vertex randomization. Nothing else. There's no subdivide happening. I made a short video below to show exactly what I mean.

Blender's documentation doesn't talk about exactly how they do the randomization - that's beyond an artist's interest. We'd have to look at the source code to see how they do it. Judging from the result though, it seems it's a randomized offset of existing vertices. So if you have one vertex offset 0.1 m in one direction, and a consecutive vertex 0.1 m in the opposite direction, then you can get a 180 degree angle.

Let's move the normals discussion to a separate issue, since it's turning out to be a longer thread. You or I can open an issue.

Let's get this PR reviewed and merged first. It's pretty minimal code change, mostly documentation. Do you have the setup to review it? If not, I could re-tag Jessica for the review.

Video of just vertex randomization on a stock cylinder object (radius 1 m, height 1 m), no other modifications:
I show normal = 1.0, then normal = 0.0:

https://user-images.githubusercontent.com/7608908/152463435-9c7d7631-b795-4dab-a53d-bed958c0e7d8.mp4

The ones I showed in the meeting for reference:

Original cylinder
2022-01-31-191548_3706x2049_scrot_cylinderSubdiv0

vertex randomization offset 0.1:
2022-01-31-194943_3706x2049_scrot_vertRand0 1

offset 0.3:
2022-01-31-194947_3706x2049_scrot

offset 0.7:
2022-01-31-194958_3706x2049_scrot

offset 1 m (same magnitude as cylinder radius and height, 1 m):
2022-01-31-195015_3706x2049_scrot_vertRand1

j-herman:
@mabelzhang I have time this morning - I'll get the review done.  I accidentally updated Blender so I can check if it works on version 3.x and downgrade my computer afterwards for the review... whoops!

mabelzhang:
> When running the tutorial as written, blender crashes (both 3.x and 2.92)

Yes! Thank you for retesting and finding that crash. I'm always looking to make my code as robust as possible, but I had not done enough checks in this prototype. Updated in e176ffd

> method variable has to be specified as an array

Updated in README and in code in e176ffd

> path, it shouldn't have a leading '/'

Are you using a relative path? I use absolute path for mine with a leading `/` and it works. I have not tried relative paths. Looks like if I do `os.getcwd()` in the Blender Python prompt, it prints the path of where the Blender executable is. That's pretty far from where DAVE is on my machine, so I've been using absolute paths.

j-herman:
Looks good at first review - I'll run it once more when I get back to the other computer and approve.  Now there's no way for me to get it wrong... :)
Yeah, I was using relative path and had just been staring at errors for too long to remember that I didn't need to do that.  Whoops.

urdf/scripts/mesh_distortion/README.md
```diff
@@ -33,22 +33,30 @@ There are two Coke can models as examples:
 `models/dave_object_models/models/Coke_Can/meshes/coke_can.obj`.
 Choose Transform while importing: Z Up, X Forward.
 
+Tip: To view the textured object in the viewport, click on the globe icon in the
```
woensug-choi, line 36: Minor suggestion,
We need not only the `obj` file but also `mtl` and  `../materials/coke.png` to load the texture correctly.

urdf/scripts/mesh_distortion/README.md
```diff
@@ -62,15 +70,17 @@ object_prefix = 'coke_can'
...
 ```
 
 Put the args into the input array:
 ```
 import sys
-sys.argv = ['distort.py', file_path, object_prefix, fouling_rating, method]
+sys.argv = ['distort.py', file_path, object_prefix, distort_extent, method]
 ```
 
 Run the script, replacing its path with the one on your machine:
```
woensug-choi, line 86: Path description in the distorted mesh's `mtl` file uses `absolute` path. The original `coke.mtl` wrote `map_Kd ../materials/coke.png
`. The distorted file produced by the distort.py script writes `map_Kd C:\\Users\\woens\\Downloads\\MeshDistortion\\Coke\\materials\\coke.png`. I think it would be better to keep to the relative.

urdf/scripts/mesh_distortion/README.md
```diff
@@ -33,22 +33,30 @@ There are two Coke can models as examples:
 `models/dave_object_models/models/Coke_Can/meshes/coke_can.obj`.
 Choose Transform while importing: Z Up, X Forward.
 
+Tip: To view the textured object in the viewport, click on the globe icon in the
+upper right of the viewport, which is for Viewport Shading: Material Preview.
+
 Look for the mesh name in the model.
```
woensug-choi, line 39: Minor suggestion,
I was trying to use custom 'object_prefix' other than 'LPCoke_Cube'. Failed, of course. The naming 'prefix' gives me a feeling that its customizable (got lazy to type 'LPCoke_Cube' and typed 'test' instead). Wouldn't it be better to change `object_prefix='LPCode_Cube'` to `object_name='LPCode_Cube'`?

woensug-choi (changes requested):
Also, I think we should change `fouling_rating` to `distort_extent` at `coke_templtae.erb` 
https://github.com/Field-Robotics-Lab/dave/blob/9eaee7f3bde44841c57511841a920e2a0db76d8e/urdf/scripts/sdf_custom_properties/coke_template.erb#L22
https://github.com/Field-Robotics-Lab/dave/blob/9eaee7f3bde44841c57511841a920e2a0db76d8e/urdf/scripts/sdf_custom_properties/coke_template.erb#L92-L93

Another minor suggestion is add a link to 2.92 blender downloads: https://download.blender.org/release/Blender2.92/

So, at the current state, a distorted mesh obj file is created when the python script for a blender is performed. Then, Ruby erb script is used to make an sdf. Could we link these procedures into one? (this may be another PR or an issue).

I've used python script and template to produce a complete set of a model object by putting all necessary files into a separate model directory with mesh, texture, and sdf when generating dissected bathymetry tiles. I may be able to reconstruct and link two procedures into one using the python console on the blender. (S-size task) This maybe useful when linking the `distort extent` to automatically calculate and write a custom SDF tag for plugins.
If you want to give it a peek about how it's done (uses linux `sed` function with `os.system`), 
- Python code block : https://github.com/Field-Robotics-Lab/Bathymetry_Converter/blob/0eeb120ac6c72d3bd28a6177987222a87a2c120a/mkbathy.py#L130-L150
- Template : https://github.com/Field-Robotics-Lab/Bathymetry_Converter/tree/master/templates

j-herman:
@mabelzhang I ran the code again and couldn't break it!  Looks good from the script perspective.  @woensug-choi made a good point about updating the demo erb and the custom_surface_properties_plugin to match the new terminology and ranges here - do you want to add those updates to this PR or are they coming separately?

urdf/scripts/mesh_distortion/README.md
mabelzhang: I added a line 5f7030b
Those files are considered part of the object model files, and it's usually assumed that the users have the entire `Coke` or `Coke_can` folder, which would include all the associated files.

Side note: I noticed that you copied some files from `Coke_Can` to a flat directory in the `nps_uw_multibeam_sonar` package, but you should really copy the entire folder, because we use relative paths inside the 3D model files to reference to the `materials`, `scripts`, etc. subfolders. Those folders also contain `model.sdf` and `model.config`, which Gazebo looks for (there's an error about missing `'model.config` for `nps_uw_multibeam_sonar/models/Coke_Can`). All models on Ignition Fuel follow this directory structure.

I would also recommend renaming the copy in `nps_uw_multibeam_sonar`, because if I do catkin install, there are two models of the same name (Coke_Can) in the GAZEBO_MODEL_PATH, and Gazebo picks one of them, such that even when I do `roslaunch dave_demo_launch distorted_coke.launch`, it picks the one in `nps_uw_multibeam_sonar` instead of the one in `dave_object_models` and gives that `model.config` error. Copying that entire folder, and renaming the folder in `nps_uw_multibeam_sonar` should disambiguate between them.

urdf/scripts/mesh_distortion/README.md
mabelzhang: Good catch. I added a `path_mode='RELATIVE'` for the OBJ export, which exposed that in the API https://docs.blender.org/api/current/bpy.ops.export_scene.html#bpy.ops.export_scene.obj 5f7030b

Side note: The COLLADA export (`bpy.ops.wm.collada_export`) though doesn't have that in the API. We aren't using that right now, as it doesn't export textures correctly, so it shouldn't affect the instructions. I added a TODO item above that line.

urdf/scripts/mesh_distortion/README.md
mabelzhang: There is a specific reason for it to be called `prefix`. If you run the script multiple times, the mesh objects will be imported with the same name, but with incremented suffix `.001`, `.002`, `.003`, etc. (I don't know why Blender does that. It does that if you do it manually too, even if you delete all the meshes in the scene.)

So we cannot just look for an exact match in the object name. We have to look for the prefix. The Python line that looks for the object uses `startswith()`.

mabelzhang:
I believe 5f7030b addresses the comments.

I debated when I opened the PR whether to rename the fouling parameter in `sdf_custom_properties`, since we weren't entirely sure whether to go the biofouling way for both the mesh modification and the Gazebo side. Now I guess we might as well. It adds a few more files to the diff - the SDF file and the Gazebo plugin.

> add a link to 2.92 blender downloads

I added the link. I also reworded the README to clarify this, because the link looks like a raw folder that users shouldn't go to. They don't necessarily need 2.92. It is not a LTS, so I don't want to recommend that specific version; it's just the version I've tested in. 2.93 is LTS but I haven't tested there. The instructions should work for newer versions. The mentioning of 2.92 is because the Blender UI changed significantly in 2.92, so item locations, shortcuts, etc. as written might not work for older versions. They should work for newer versions though.

> Could we link these procedures into one?

I don't know if that's possible, because the Blender Python script is executed from a Python interpreter within the Blender GUI, whereas the Ruby script is run on the command line, outside Python.

I also don't know whether it's desirable to couple the two scripts. They take different parameters. The mesh distortion script is an artistic process for a general model file, whereas the SDF Ruby template is a physical process specific to parameters that will affect results in Gazebo.

We happen to have the biofouling rating in both, but you can see they weren't the same name (now they are, but) as the two scripts weren't intended to be coupled. Somebody could add any custom SDF tags to their file, like material properties, with or without distortion. The fouling / distort SDF tag is just an example to illustrate the users can add whatever they want. If the user doesn't have any distortion, then there's really no relation between the two scripts to warrant linking them.

Right now, it doesn't take too much effort for us to put in the same distort value for both. I don't know if it'll be worth the effort to link them. I also think that's something we can leave up to the user to do, as opposed to handling in DAVE.

Less relevant, but when we upstream these two scripts, they will live in different repos. One will be in an SDF repo, and the other will be in `ign-common`, precisely because they do such separate things.

urdf/scripts/mesh_distortion/README.md
woensug-choi: Gotcha

urdf/scripts/mesh_distortion/README.md
woensug-choi: Understood. I should learn `startswith()`

woensug-choi (approved):

j-herman (approved):

woensug-choi:
FYI, @mabelzhang I've updated the multibeam sonar repo for the duplicate model directory. It's not in SDF file structure but a bare OBJ and MTL files now in the `coke_can_example` directory. I think it's better for the sonar tutorial perspective that all SDF custom tags can be seen in a single `world` file to play with.

PR #189, opened 2022-02-08
Elec lead

dtdavi1:
Updated flying lead functionality can be tested with either of the following demos (only the second includes the new plug/receptacle model):
```
ros_launch dave_demo_launch dave_electrical_mate.launch
```
or
```
roslaunch dave_demo_launch dave_plug_and_socket_demo.launch
```

Updates include the following:
Force calculations for insert & extract aligned with the plug link
New meshes and models
     Male-to-female and female-to-male plug/receptacle combos
     Underwater platform model (incorporated into both SDF and URDF demos)
Added ROS functionality to expose calculated force applied to the plug link during insert & extract
    ```rqt_plot /m_to_f_plug/appliedForce/vector``` to visualize the message
Tutorial updates to account for changes

### DIFF
--- a/models/dave_object_models/models/f_to_m_receptacle/model.sdf
+++ b/models/dave_object_models/models/f_to_m_receptacle/model.sdf
@@ -10,7 +10,7 @@
           <mesh>
             <uri>model://f_to_m_receptacle/meshes/Plug-Receptacle.dae</uri>
             <!-- Smaller scale than the plug to make the fit more forgiving -->
-            <scale>0.375 0.375 0.375</scale>
+            <scale>0.365 0.375 0.375</scale>
 
           </mesh>
         </geometry>
@@ -20,7 +20,7 @@
         <geometry>
           <mesh>
             <uri>model://f_to_m_receptacle/meshes/COLLISION-Plug-Receptacle.dae</uri>
-            <scale>0.375 0.375 0.375</scale>
+            <scale>0.365 0.375 0.375</scale>
           </mesh>
         </geometry>
       </collision>
@@ -51,8 +51,8 @@
       <plugModel>f_to_m_plug</plugModel>
       <plugLink>f_to_m_plug</plugLink>
       <rollAlignmentTolerance>0.3</rollAlignmentTolerance>
-      <pitchAlignmentTolerance>0.3</pitchAlignmentTolerance>
-      <yawAlignmentTolerance>0.3</yawAlignmentTolerance>
+      <pitchAlignmentTolerance>0.1</pitchAlignmentTolerance>
+      <yawAlignmentTolerance>0.1</yawAlignmentTolerance>
     </plugin>
 
   </model>
--- a/urdf/manipulators/plug_and_socket_description/urdf/f_to_m_receptacle.xacro
+++ b/urdf/manipulators/plug_and_socket_description/urdf/f_to_m_receptacle.xacro
@@ -18,14 +18,14 @@
         <geometry>
           <!-- Smaller scale than plug to make the fit more forgiving -->
           <mesh filename="package://dave_object_models/models/f_to_m_receptacle/meshes/Plug-Receptacle.dae"
-                scale="0.375 0.375 0.375"/>
+                scale="0.365 0.375 0.375"/>
         </geometry>
         <origin xyz="0 0 0" rpy="${pi/2} 0 ${pi/2}"/>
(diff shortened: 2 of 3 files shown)


bsb808:
Is the example `roslaunch dave_demo_launch dave_plug_and_socket_demo.launch` represented in the wiki pages somewhere?    I see `roslaunch dave_demo_launch dave_electrical_mate.launch` in https://github.com/Field-Robotics-Lab/dave/wiki/Electrical-Plug-Mating-Plugin.   

Maybe we could add the first roslaunch to the existing wiki, add a screenshot and explain the differences between the two examples?

bsb808:
I was able to run both roslaunch examples and see that the `/m_to_f_plug/appliedForce/vector` was advertised.  I'm probably not skilled enough to drive the vehicle and arm to test this, so my proposed plan is...

1. Once the documentation (see comment above) is addressed, then I'll approve.
2. Go ahead and merge
3. Find someone to take on https://github.com/Field-Robotics-Lab/dave/issues/193
4. Once we have solved https://github.com/Field-Robotics-Lab/dave/issues/193, use that to test/demonstrate the mating forces and create a highlight video.

@dtdavi1 Sound like a plan?

dtdavi1:
I've updated the wiki some to better differentiate between the two demos (the dave_plug_and_socket_demo is towards the end).  I'm working on a video demo and will finalize that before the release.

dtdavi1:
For a quick test of the plugin functionality, the [world file](https://github.com/Field-Robotics-Lab/dave/blob/elec_lead/models/dave_worlds/worlds/dave_electrical_mating.world) that is used in the dave_electrical_mate.launch demo can be modified as follows:
Comment out line 174
Uncomment line 175

This will position the plug in the socket's opening (but not engaged) upon startup.  The prismatic joint will be formed right after startup, but there won't be enough applied force to lock it.  If you maneuver the RexROV to the stand and lightly bump the plug downward, the joint should lock almost immediately (and should be indicated in the terminal window).

Unlocking will still require you to grab the plug handle with the gripper and pulling upwards (but that's much easier than doing the whole insertion evolution).

bsb808:
@dtdavi1 Both of the launch examples run for me.

I ran the example you suggest to start with the plug aligned.  When I nudge it with the ROV I get some errors reported to the terminal:

```
Msg] m_to_f_socket and m_to_f_plug are aligned in orientation and distance
[Msg] m_to_f_socket and m_to_f_plug within proximity
[Msg] m_to_f_socket-m_to_f_plug joint formed
[Msg] m_to_f_socket-m_to_f_plug joint locked!
Error [Param.cc:457] Out of range. Unable to set value [6.94416e-310 ] for key[upper].
Error [Param.cc:457] Out of range. Unable to set value [6.94416e-310 ] for key[lower].
[Msg] m_to_f_socket-m_to_f_plug joint unlocked!
[Msg] m_to_f_socket-m_to_f_plug joint removed
[Msg] m_to_f_socket and m_to_f_plug are aligned in orientation and distance
```

Are these expected?

dtdavi1:
@bsb808 They're all either expected or at least not surprising.  Here's a summary of what the stream is telling you:
The first is a report that the plug and socket are properly aligned.
The second is a report that the prismatic joint was created (i.e., they've been properly aligned for the requisite amount of time)
The third is a report that the joint was locked (i.e., it's plugged in) because the right amount of pressure was applied
I have no idea what the errors are (they occur on occasion but don't really affect anything).  You can write it up as an issue (or I will) if you want, but I don't think they should prevent the merge.
The "joint unlocked" and "joint removed" mean that it was unplugged.  If you didn't actually grab it and pull, it was probably a byproduct of the plug being jostled when it was plugged in (i.e., it plugged in successfully but then popped loose).
The last line is from the process starting all over because it's not plugged in but is properly aligned.

bsb808:
@dtdavi1 Understood.  The [Msg]s are really helpful and I understood those - it was the Error that makes me unsure if things are operating as expected.

I opened an issue so we can run this to ground post 4.3.0 - https://github.com/Field-Robotics-Lab/dave/issues/227

Please merge when you can.

bsb808 (approved):
Ready to merge

dtdavi1:
@bsb808 Sounds good.  My best guess is that the error is associated with a race condition that occurs when the joint is locked and then removed too quickly (which occurs somtimes when the insertion is rough).

PR #194, opened 2022-02-11
New models

bsb808:
This PR adds three new object models (torpedo, flight-data-recorder, sonobuoy and three unexploded ordnance) along with a demonstration world.

Reviewer should be able to run the command 
```
```
and view a minimal world with the three now models on the seafloor: 

Screenshot from 2022-02-11 16-34-37

Please check that the link coordinate frames make sense and that the collision meshes are appropriate.  I find it easiest to use the Gazebo gui to view this attributes.

### DIFF
--- a/models/dave_object_models/models/flight_data_recorder/inertial_calcs.py
+++ b/models/dave_object_models/models/flight_data_recorder/inertial_calcs.py
@@ -1,0 +1,16 @@
+'''
+iPython script to calculate inertial properties
+'''
+from math import pi
+
+# Mass [kg]
+m = 1.0
+# Length and dia [m]
+l = 0.5
+# "mass" in water [kg]
+m_water = m 
+print("Mass in water %f kg"%m_water)
+
+ixx = 1.0/6.0*m*l**2
+
+print("ixx = iyy = izz = %f"%(ixx))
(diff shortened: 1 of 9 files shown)


mabelzhang (commented):
Someone else who knows what's "correct" should review this. I just took a brief look to steal the SDF file to test other models.

Coordinate frames look right - assuming all of them being Z up X forward is what we want.
2022-02-12-035143_3706x2049_scrot

Inertia boxes look off. Do we care about that? If we are planning to use them for manipulation, that might not be negligible.
2022-02-12-035234_3706x2049_scrot

Side note, I see the light/dark shadow upstream bug now... Guess I have the dark version.

Also saw another upstream bug that if I enable skeleton view, then disable it, then enable inertia view, all the magenta boxes are opaque as opposed to translucent. Probably not an urgent thing to fix. FYI @WilliamLewww

bsb808:
@mabelzhang Good catch on the inertia values.  I improved them.  Still coarse estimates, but I think they should be sufficient for dave.
Screenshot from 2022-02-18 17-52-32

mabelzhang (approved):
Oh never mind, that one was already correct.

Thanks for adding them!

PR #196, opened 2022-02-12
Fix catkin build installs for dave_demo.launch

mabelzhang:
## Premise
Since #148, I've been adding missing lines to `CMakeLists.txt` files when I catch them, so that they work for the stricter `catkin build`, and moreover, `catkin config --install`. The purpose is to improve the quality of the software so that CMake lines are more correct, and packages can be installed and still run correctly.

The repo's default instructions recommend using `catkin_make`, which is more lenient and doesn't complain about things even if they're incorrect.

## Fixes
The `catkin_package()` fixes the "Resource not found" error. That is required for `rospack find` to find a package.
The `install()` fixes the "No such file or directory" error.

## To test

Build with `catkin build` (as opposed to `catkin_make`), with the `install` option:
```
catkin config --install
catkin build
. install/setup.bash
```

Run this demo file
```
roslaunch dave_demo_launch dave_demo.launch
```

Without this PR, you'll get errors about resource not found and files in `urdf/` not exist:
```
Resource not found: rexrov_description
```
```
RLException: while processing /home/master/dave_ws/install/share/rexrov_description/launch/upload_rexrov_oberon7.launch:
Invalid roslaunch XML syntax: [Errno 2] No such file or directory: '/home/master/dave_ws/install/share/rexrov_description/launch/upload_rexrov_oberon7.launch'
```
```
resource not found: dave_snippets
```
```
No such file or directory: /home/master/dave_ws/install/share/dave_snippets/urdf/sensor_snippets.xacro [Errno 2] No such file or directory: '/home/master/dave_ws/install/share/dave_snippets/urdf/sensor_snippets.xacro'
```
(Similarly for `dave_oberon7_description`)

With this PR, Gazebo GUI should show up, and there are no such errors.

Test that this didn't break the `catkin_make` build:
```
# First, (re)move the built files from above. Then rebuild from scratch:
catkin_make
. devel/setup.bash
roslaunch dave_demo_launch dave_demo.launch
```

### DIFF
--- a/urdf/dave_snippets/CMakeLists.txt
+++ b/urdf/dave_snippets/CMakeLists.txt
@@ -3,9 +3,13 @@ project(dave_snippets)
 
 find_package(catkin REQUIRED)
 
+catkin_package()
+
 include_directories(
 # include
 # ${catkin_INCLUDE_DIRS}
 )
 
-
+install(DIRECTORY urdf
+        DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+        PATTERN "*~" EXCLUDE)
--- a/urdf/manipulators/dave_oberon7_description/CMakeLists.txt
+++ b/urdf/manipulators/dave_oberon7_description/CMakeLists.txt
@@ -3,9 +3,13 @@ project(dave_oberon7_description)
 
 find_package(catkin REQUIRED)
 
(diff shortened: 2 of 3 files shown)


Yadunund (approved):
LGTM!

PR #199, opened 2022-02-14
Improve CI

senceryazici:
Hi, 
I've configured ROS `industrial_ci` (which is almost standard for building ros workspaces) for you to build `dave`. Also added cmake args for building with `lld` as linkerscript (faster than ld).

Please note that, even though the `dave` repository is included in [dave_sim.repos](https://github.com/Field-Robotics-Lab/dave/blob/master/extras/repos/dave_sim.repos) (causing CI to build & test it twice), it builds faster due to `lld`. `industrial_ci` will build `dave` as `upstream_ws` first (since it's in [dave_sim.repos](https://github.com/Field-Robotics-Lab/dave/blob/master/extras/repos/dave_sim.repos) ), and then will build as `target_ws` since it's the actual repository, I have not removed `dave` repository from [dave_sim.repos](https://github.com/Field-Robotics-Lab/dave/blob/master/extras/repos/dave_sim.repos). You can either remove it from there (preffered), or create another `.repos` files **without** `dave`, to make CI even faster (about 17 mins.)

Also, `.repos` files with ssh links ( such as `git@github:...`) will fail in `industrial_ci`. I've added a sed command to replace them with `https` [in this line](https://github.com/senceryazici/dave/blob/6069d32ed7ebc035d7f022641af9589c3327b73f/.github/workflows/ci.yml#L34).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
         -DCMAKE_SHARED_LINKER_FLAGS=-fuse-ld=lld
         -DCMAKE_MODULE_LINKER_FLAGS=-fuse-ld=lld
         -DCMAKE_BUILD_TYPE=Release
+      ADDITIONAL_DEBS: lld python3-catkin-tools python3-osrf-pycommon
     runs-on: ubuntu-20.04
 
     steps:


mabelzhang:
I edited the post so that they point to the correct URL for the `dave_sim.repos` file.

M1chaelM:
@senceryazici Thanks for doing this! It looks like an improvement to me and I'm ready to try it out, but I have a question about your change to the code_check.sh script. I leave you a comment inline.

tools/code_check.sh
```diff
@@ -156,3 +156,4 @@ if [ $xmlout -eq 1 ]; then
 elif [ $QUICK_CHECK -eq 0 ]; then
   echo $CPPLINT_FILES | xargs python tools/cpplint.py 2>&1
 fi
+echo ""
```
M1chaelM, line 159: @senceryazici What does this addition do?

M1chaelM: Since it is definitely harmless, I'm going to go ahead and merge, but if you can explain I'd appreciate it.

M1chaelM (approved):
This looks good, and we're going to try it out. It sounds like removing the dave repos from dave_sim.repos would ultimately make sense also, though we will do that later just for logistical reasons relating to our upcoming release.

tools/code_check.sh
senceryazici: While testing on my local with 'act' (a tool for running GitHub actions on your local) the code check script constantly failed for some reason. Just for debugging I added a line to the end to see the result and it started passing.

Honestly (just like you've said) since it's totally harmless I didn't spend extra time on it's reason and let it be that way. But I'll definitely take a look at it when I have available time.

bsb808:
> This looks good, and we're going to try it out. It sounds like removing the dave repos from dave_sim.repos would ultimately make sense also, though we will do that later just for logistical reasons relating to our upcoming release.

@M1chaelM If you haven't done it yet, could you post an issue to the tracker to describe what needs to be done so that we have it on our list for post release?

Issue #191, opened 2022-02-11
Write a tutorial on using ignition fuel models


bsb808:
Maybe this already exists externally, but now that we have our [DAVE collection](https://app.ignitionrobotics.org/Cole/fuel/collections/DAVE), it would be cool to have a wiki tutorial illustrating how to use the models from the collection.

mabelzhang:
We have a pretty comprehensive tutorial on the Ignition website here https://ignitionrobotics.org/api/gazebo/6.5/meshtofuel.html

The wiki page should at a minimum link to that, as it gives information about how the model directory should be structured, which is useful for uploading models. It also gives an example of how to use the model in a world file - it's literally a one-step process of copy-pasting the URI on the webpage and throwing it into the SDF `<include><uri>`.

PR #205, opened 2022-02-17
Add new models from Ignition Fuel

mabelzhang:
Builds on top of #194
Demonstrates an example for #191

Adds new models (`Lionfish`, `Niskin`, `Sunken Vase`, `Sunken Ship`) Cole uploaded to the [DAVE collection](https://app.ignitionrobotics.org/Cole/fuel/collections/DAVE) on Ignition Fuel.

They took a few iterations of troubleshooting. `Sunken Vase 02` is commented out for now because it throws an OGRE error, and we're not sure why.

Inertial values are from MeshLab. I'm not planning to add inertia for the ship and the fish, since they're mostly for decoration and maybe for sensor scanning, not any kind of physical interaction. (At least... I wasn't told we were planning to do any deep sea fishing.)

To test:
```
```

You should see messages printed about downloading the models from Ignition Fuel:
```
[Msg] Downloaded model URL: 
          https://fuel.ignitionrobotics.org/1.0/Cole/models/Sunken Vase
      to: 
...
```

They should show up like this (3 objects on the far right and the big ship):
Here I started Gazebo paused to show their initial poses. When unpaused, they fall to the ground and won't look upright.
2022-02-19-022811_1955x1183_scrot_SunkenShip
2022-02-19-022819_1955x1183_scrot

Inertia for Sunken Vase and Niskin:
2022-02-19-053157_3706x2049_scrot

### DIFF
--- a/models/dave_object_models/models/flight_data_recorder/inertial_calcs.py
+++ b/models/dave_object_models/models/flight_data_recorder/inertial_calcs.py
@@ -1,0 +1,16 @@
+'''
+iPython script to calculate inertial properties
+'''
+from math import pi
+
+# Mass [kg]
+m = 1.0
+# Length and dia [m]
+l = 0.5
+# "mass" in water [kg]
+m_water = m 
+print("Mass in water %f kg"%m_water)
+
+ixx = 1.0/6.0*m*l**2
+
+print("ixx = iyy = izz = %f"%(ixx))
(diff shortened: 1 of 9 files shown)


woensug-choi:
image

Called in just fine 👍

woensug-choi (approved):
Fine-tuned collision model!

mabelzhang:
The collision models are currently just the mesh itself. We could make them into boxes later if that turns out to affect manipulation performance!

PR #212, opened 2022-02-19
Degraded models and degradation script update

mabelzhang:
Goes with https://github.com/Field-Robotics-Lab/nps_uw_multibeam_sonar/pull/33

Adds
- Distorted Sunken Ship (only mesh, model.sdf. model.config). Other files use upstream ones on Ignition Fuel
- Distorted Sunken Vase. Same situation as above.

Modified `VERT_RAND_MAX` in mesh distortion script to work better for larger objects than the Coke can.

There isn't a world/launch file in this PR itself, as the models will eventually go to the world file in #194.
Here's a visualization of the distorted models next to the undistorted originals:
2022-02-19-061811_3706x2049_scrot
2022-02-19-080508_3706x2049_scrot

### DIFF
--- a/examples/dave_demo_launch/launch/dave_demo.launch
+++ b/examples/dave_demo_launch/launch/dave_demo.launch
@@ -8,7 +8,7 @@
     <arg name="timeout" default="0.0"/>
     <arg name="velocity_control" default="true"/>
     <arg name="joy_id" default="0"/>
- 
+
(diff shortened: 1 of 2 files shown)


woensug-choi (approved):
Made some changes for the model url typo fix. Other than that, models are being called just fine.

mabelzhang:
Ooh good catch! I have two copies of that folder on my local machine, ha.

PR #214, opened 2022-02-21
Fix out of bounds read in ocean current world plugin

daewok:
Found using valgrind


daewok:
Found another issue using valgrind. Going to open a new PR for everything.

PR #216, opened 2022-02-21
Fix reading from uninitialized variables

daewok:
There were several places where behavior of the Gazebo plugins depended on uninitialized memory. This has bitten me in the past (Field-Robotics-Lab/glider_hybrid_whoi#63), so explicitly initialize everything where possible. (And don't read beyond the end of a vector).

These issues were find using valgrind and the glider_hybrid_whoi stack.


woensug-choi (approved):
Thanks for picking up the crumbles...

daewok:
No worries. Now I know how to use valgrind :)

PR #219, opened 2022-02-22
Scale integrated world seabed

Yadunund:
After some consideration, I felt the current seabed is too vast and hence I've scaled it down to something more manoeuvrable by the UUV. At the same time, i've scaled it anisotropically to get some steep cliffs which might make the terrain more interesting.

output

The UUV now spawns above the central mound-like feature. I'd like to propose we setup demo stations around this mound as seen below. (use the direction of reference axes to determine where these spots are)

image

### DIFF
--- a/models/dave_worlds/worlds/dave_integrated.world
+++ b/models/dave_worlds/worlds/dave_integrated.world
@@ -160,6 +160,50 @@
 
     <plugin name="sc_interface" filename="libuuv_sc_ros_interface_plugin.so"/>
 
+    <!-- Electrical flying lead -->
+    <include>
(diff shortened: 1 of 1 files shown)


quarkytale (approved):
Looks good to me!

quarkytale:
A suggestion on demo stations locations to be surrounding the vehicle's initial position in a circular or square fashion.

Yadunund:
> A suggestion on demo stations locations to be surrounding the vehicle's initial position in a circular or square fashion.

Yes that is the idea here. Going with a circular pattern wherever the topology allows. As we introduce more stations, we can distribute them radially.

PR #217, opened 2022-02-22
Integrated World

Yadunund:
An underwater world that showcases various simulator capabilities. See Usage below.

## Pending items:
- [x] Select seabed model. 
  - [x] Currently, this branch has its own copy of the Santorini mode. TODO: Use the model from this PR once it is merged https://github.com/Field-Robotics-Lab/dave/pull/203.
    - Created a new `SantoriniScaled` model that imports the mesh from the original `Santorini` model https://github.com/Field-Robotics-Lab/dave/pull/217/commits/98a6b895f27ad1ee09e5a8a065b9041b262749b0 https://github.com/Field-Robotics-Lab/dave/commit/98a6b895f27ad1ee09e5a8a065b9041b262749b0
  - [x] Consider further scaling the seabed. The current scaling is `0.1`
    - Scaled X & Y to `0.05`. Z remains at `0.1`. https://github.com/Field-Robotics-Lab/dave/pull/219
- [x] Incorporate electrical flying lead https://github.com/Field-Robotics-Lab/dave/pull/223
- [x] Incorporate a mud pit with objects to drag around https://github.com/Field-Robotics-Lab/dave/pull/221
- [x] Incorporate bimanual demo https://github.com/Field-Robotics-Lab/dave/pull/240/
- [x] Script to teleport UUV to various stations/demo sites https://github.com/Field-Robotics-Lab/dave/pull/231
- [x] Add models from https://github.com/Field-Robotics-Lab/dave/pull/242
  - [x] https://app.ignitionrobotics.org/search;q=dave
  - [x] https://github.com/Field-Robotics-Lab/dave/pull/194
  - [x] https://github.com/Field-Robotics-Lab/dave/pull/209
- ~[ ] [Optional] Incorporate multibeam demo https://github.com/Field-Robotics-Lab/dave/pull/222~
- [x] [Optional] Incorporate multiple UUVs with separate joystick control https://github.com/Field-Robotics-Lab/dave/pull/240/

image

## Usage
To launch the demo world
```bash
roslaunch dave_demo_launch dave_integrated_demo.launch
```
## UUVs
The world spawns two UUVs, a single arm `rexrov0` UUV which can be teloperated using a joystick connected to `js0` and a dual arm `rexrov` UUV which can be teleoperted using a joystick connected to `js1`. The arms on this UUV can be commanded via the moveit interface. See Electrical Station below for more information.

## Stations
[...]

PR #221, opened 2022-02-24
Sand occlusion and mudpit in integrated world

Yadunund:
This PR adds visual silt occulision along with a mudpit with appropriate plugins. 

image

```bash
roslaunch dave_demo_launch dave_integrated_demo.launch 
```

### DIFF
--- a/models/dave_worlds/worlds/dave_integrated.world
+++ b/models/dave_worlds/worlds/dave_integrated.world
@@ -234,28 +234,28 @@
             </contact>
           </surface>
         </collision>
-      <visual name='visual'>
-        <geometry>
-          <mesh><uri>model://SantoriniScaled/meshes/mud.dae</uri></mesh>
-        </geometry>
-        <!-- <transparency>0.1</transparency>
-        <material>
-          <diffuse>0.1 0.075 0.05 1.0</diffuse>
-          <emissive>0.02 0.02 0.02 1.0</emissive>
-        </material> -->
-        <material>
-          <script>
-            <uri>model://occluded_sand_heightmap/materials/mud.material</uri>
-            <name>Mud</name>
-          </script>
-          <emissive>
-            0.6 0.6 0.6 1.0
-          </emissive>
-        </material>
-      </visual>
+        <visual name='visual'>
+          <geometry>
+            <mesh><uri>model://SantoriniScaled/meshes/mud.dae</uri></mesh>
+          </geometry>
+          <!-- <transparency>0.1</transparency>
+          <material>
+            <diffuse>0.1 0.075 0.05 1.0</diffuse>
+            <emissive>0.02 0.02 0.02 1.0</emissive>
+          </material> -->
+          <material>
+            <script>
+              <uri>model://occluded_sand_heightmap/materials/mud.material</uri>
+              <name>Mud</name>
+            </script>
+            <emissive>
+              0.6 0.6 0.6 1.0
+            </emissive>
+          </material>
+        </visual>
         <sensor name="mud_contact" type="contact">
           <always_on>true</always_on>
-          <update_rate>1000</update_rate>
+          <update_rate>1</update_rate>
           <contact>
             <collision>collision</collision>
           </contact>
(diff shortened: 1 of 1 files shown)


j-herman:
@Yadunund The world loads up with no problem and I spent some time trying to drag the paddle around.  It's challenging!  Possibly because of my limited joystick skills...
Anyhow, this works fine so I'm happy to approve the PR, but there are a few things that I think would be good to consider for updates.
One thing I noticed is that I kept losing the paddle in the mud, where it gets deep.  For the original demo, I made the mud pit follow the contours of the terrain - a layer, rather than a pool.  Another way to do that would be to put a solid surface underneath the mud model.  What do you think?
Also, it would be fun to have other objects to play with - maybe the vase, since it looks like it has a good handle that the manipulator might be able to grasp?  I was thinking about adding something else into the basic demo but maybe this would be a better playground.

Yadunund:
Hi @j-herman 

Thanks for trying this out so quickly!

> I spent some time trying to drag the paddle around. It's challenging!

I did tripe the damping from your original demo so I wonder if that's making things harder? I can reduce it if so. Another factor could be the large default inertia values of the anchor object. Perhaps we can correct these values in `master` and we can merge those changes in here?
image

>  For the original demo, I made the mud pit follow the contours of the terrain - a layer, rather than a pool. Another way to do that would be to put a solid surface underneath the mud model. What do you think?

Ah that makes a lot of sense! I'm happy to add another surface underneath. I went with the a flat pool so that the terrain would naturally give rise to shallower and deeper regions (mimicking reality?) . Another option is to lower the pose of the mud mesh. I think you have better expertise on this matter so let me know what's best and i'll update accordingly. 

> Also, it would be fun to have other objects to play with - maybe the vase, since it looks like it has a good handle that the manipulator might be able to grasp? 

This is a great suggestion. My plan was to [wait till the end](https://github.com/Field-Robotics-Lab/dave/pull/217#issue-1146325867) before populating the world with additional models. There are some nice vases in the fuel collection so I'll try adding those in

j-herman:
The hardest part for me is trying to get the gripper around the handle, which is a function of the position of the paddle.  My joystick hasn't been working well for some of the arm/wrist degrees of freedom so anything on the ground is almost impossible to grab.  Also, I kept pushing the paddle into the mud pit and losing it altogether.  Like I said, possibly just my own lack of skills :)

I did notice you'd increased the damping.  Actually, both that and the significant inertia seem to make it easier to grasp the handle, since it doesn't fall over as easily.  I did try to add in some realistic inertia values when I first built the anchor, but the results were not great, so I took all of that back out since the paddle was just a demonstration tool.  Definitely give it a shot if you think it will help.  
I started with damping values between 1 and 10 after looking at the original mud demo world, and then doubled it to make the effects more obvious.  I haven't been able to find a good reference that would help us convert damping to material viscosity in some physically accurate way, so our method for choosing values is still just based on producing behavior that seems reasonable.
I do like the way the variable depth looks and works, so maybe there is a way to keep that while still limiting the lost-in-the-muck trouble.  We'd be simulating two different seafloor types - a muddy layer of moderately consistent thickness, and something more like a pool that objects would sink into fully and get lost.  There may be an easy way to warp a section of the seafloor in Blender to generate that effect.  Scaling the z coordinates, possibly?

Yadunund:
@j-herman sorry for the delay but this is ready for another review. I've split up the mud pit into two as seen in the image below. The one in the foreground is a pool with variable depth while the other is constant depth that follows the terrain.

image

j-herman:
@Yadunund I like the idea - looks good!  I've got a connectivity issue that will hopefully be resolved on Thursday, so will plan to review this on Friday.  
What are your thoughts on the inertia values for the mud anchor?  If you think it would be worth trying I can put the realistic values back in, and we can see how it goes.

Yadunund:
@j-herman thanks! I leaning towards leaving the inertias as they are for now. Perhaps in the future we can experiment with what works.

j-herman (approved):

PR #223, opened 2022-02-24
Adding electrical mating current demo

quarkytale:
Confirm the location for this demo station.
new_panel
`roslaunch dave_demo_launch dave_integrated_demo.launch `

### DIFF
--- a/models/dave_worlds/worlds/dave_integrated.world
+++ b/models/dave_worlds/worlds/dave_integrated.world
@@ -160,15 +160,15 @@
 
     <plugin name="sc_interface" filename="libuuv_sc_ros_interface_plugin.so"/>
 
-        <!-- Electrical flying lead -->
+    <!-- Electrical flying lead -->
     <include>
       <uri>model://m_to_f_plug</uri>
-      <pose>-190.5 117 -155.89 -1.5707963267948966 0</pose>
+      <pose>-190.44 117.25 -155.84 0 -1.570777 0</pose>
     </include>
-  
+
     <include>
       <uri>model://f_to_m_plug</uri>
-      <pose>-190.5 118 -155.95 0 -1.5707963267948966 0</pose>
+      <pose>-190.5 118 -155.84 0 -1.570777 0</pose>
     </include>
 
     <model name="platform">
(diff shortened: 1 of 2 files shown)


models/dave_worlds/worlds/dave_integrated.world
```diff
@@ -160,49 +160,55 @@
 
     <plugin name="sc_interface" filename="libuuv_sc_ros_interface_plugin.so"/>
 
-    <!-- Electrical flying lead -->
+        <!-- Electrical flying lead -->
```
Yadunund, line 163: Undo indent for comment

models/dave_worlds/worlds/dave_integrated.world
```diff
@@ -160,49 +160,55 @@
...
-      <entity>wamv::base_link</entity>
-      <namespace>vrx/dock_2018/bay_1_external</namespace>
-      <pose frame="robotx_dock_2018::dock_2018_placard1::placard::link">35 -10 -110 0 0 0</pose>
-      <geometry>
-        <box>
-          <size>8 1.5 2</size>
-        </box>
-      </geometry>
-    </plugin>
+    <include>
```
Yadunund, line 207: The `electrical_panel` model only has one set of receptacle and plug. But we should include the latest panel that is featured in the `plug_and_socket_demo` which has multiple sets. This is done by invoking a xacro and spawing the resulting model. Let's adopt the same approach instead. So basically you'll need to copy [this block](https://github.com/Field-Robotics-Lab/dave/blob/3507bccc51e214aa825b0b16651e9cd7de18814a/examples/dave_demo_launch/launch/dave_plug_and_socket_demo.launch#L47-L49) into the launch file instead of including the electrical_panel model in the world file.

Yadunund (changes requested):
Thanks for getting started on this.

Apart from the highlighted comments, please reposition the platform and panels so that they do not appear to be floating. This is what it currently looks like 
image

This is what it should look like as seen in the `dave_plug_and_socket` demo 
image

quarkytale:
Thanks for checking it out @Yadunund, can you check in the GUI if the coordinates of the panel are same as in the world file? I was getting it slightly skewed a couple of times during my tests.

Yadunund:
Since you didn't include the panel as a `static` model, it remains in collision with the seabed and hence gets pushed up. I've fixed that along with some other tweaks in this commit https://github.com/Field-Robotics-Lab/dave/pull/223/commits/ea70b8d3fdd7cdf6022ead1e22d11c3e8cf95117

Yadunund (approved):

PR #209, opened 2022-02-19
More new models.

bsb808:
Extends PR #194 by adding 4 hardhat (Benthos hardhat floats) and MBARI MARS node models.  Recommend reviewing and merging after #194 is merged.

Reviewer should be able to run the command

```
roslaunch dave_demo_launch dave_models.launch
```
 
and view a minimal world with the new models on the seafloor:

Screenshot from 2022-02-18 19-31-12

Issue #220, opened 2022-02-23
Add a wiki page as tutorial/demo of the new models


bsb808:
See PR #209 

* Still need to add airplane wreckage.
* Mention that it can take some time b/c of fuel download and link to fuel page.
* List the models with short descriptions so google can find them.

bsb808:
Completed with https://github.com/Field-Robotics-Lab/dave/wiki/Dave-Object-Models

PR #225, opened 2022-02-25
Adding heavy weight torpedo model

bsb808:
Adds one more model - the Mark 48 heavy weight torpedo.

To test run
```
```

You should see one additional model compared to https://github.com/Field-Robotics-Lab/dave/pull/209 - the large torpedo shown here:

Screenshot from 2022-02-24 20-36-00

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,70 +1,38 @@
 name: Ubuntu CI
 
-on: [pull_request]
+on: 
+  workflow_dispatch:
+  pull_request:
+
 
 jobs:
-  Test-Focal-Noetic-Gazebo11:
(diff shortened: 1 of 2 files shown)


bsb808:
A tagged both @quarkytale and @j-herman as potential reviewers.  Just needs one review, but thought I'd see who has the time to take a look.

j-herman:
I've got it.  Nice easy way to ease into the morning :)

j-herman (approved):
Looks like a torpedo.  When lifted and dropped, rolls like a big heavy torpedo.  All good here!  Definitely a good call to mention the extended load time in the wiki as you mentioned in https://github.com/Field-Robotics-Lab/dave/issues/220 - it surprised me.

PR #228, opened 2022-02-25
Simple bimanual demo

crvogt:
Simple example node that demonstrates some of the rexrov dual-arm capabilities.

To run, start rexrov with dual oberon7s with
```roslaunch dave_demo_launch dave_oberon7_moveit.launch```

Once the previous command has finished loading, start RViz and launch `move_group` with 
```roslaunch rexrov_oberon7_moveit rexrov_dual_arm_moveit_planning_execution.launch moveit_controller_manager:=rexrov```

Once RViz has started, move to the robot in Gazebo or check RViz to view demo which is run with
```rosrun dave_nodes bimanual_simple_demo.py```

The sequence should appear as
- open and close each gripper
-  move each end effector to a goal pose
- each arm returns to a start pose

### DIFF
--- a/examples/dave_nodes/src/bimanual_simple_demo.py
+++ b/examples/dave_nodes/src/bimanual_simple_demo.py
@@ -329,9 +329,8 @@ class MoveGroupPythonInterface(object):
 
         return new_target_pose
 
-    def run_node(self):
+    def run_node(self, action_num):
         # Create a counter to give sim time to give object coords
-        counter = 0
         while not rospy.is_shutdown():
             # Get the location of a target relative to the robot in 
             # arm command coordinates
@@ -340,23 +339,30 @@ class MoveGroupPythonInterface(object):
                                                self.get_state_msg.pose[-1],
                                                self.move_group_arm_l.get_current_pose().pose)
 
-            if counter == 2:
-                try:
-                    # Open and close each gripper
+            try:
+                # Open and close each gripper
+                if action_num == 0:
                     self.open_gripper_r()
+                elif action_num == 1:
                     self.close_gripper_r()
+                elif action_num == 2:    
                     self.open_gripper_l()
+                elif action_num == 3:
                     self.close_gripper_l()
+                elif action_num == 4:    
                     # Go to first pose goal, set home pose as False
                     self.go_to_pose_goal_r(False)
+                elif action_num == 5:    
                     self.go_to_pose_goal_l(False)
+                elif action_num == 6:    
                     # Set home pose as true to return to start
                     self.go_to_pose_goal_r(True)
+                elif action_num == 7:    
                     self.go_to_pose_goal_l(True)
-                except:
-                    print("Unable to assume pose")
-                break
-            counter = counter + 1
+            except:
+                print("Unable to assume pose")
+            break
+
             self.rate.sleep()
 
 
@@ -364,8 +370,8 @@ def main():
     try:
         print("Setting up moveit commander")
         bimanual_demo = MoveGroupPythonInterface()
-
-        bimanual_demo.run_node()
+        for action_num in range(8):
+            bimanual_demo.run_node(action_num)
 
     except rospy.ROSInterruptException:
         return

### CODE REVIEW

**Comment by mabelzhang** [comment]
I'm getting a `No such file or directory` for `oberon7_description/urdf/common.gazebo.xacro`

I think it comes from this line

https://github.com/Field-Robotics-Lab/dave/blob/ef45e67ee4782830023880e757b853e9cdbd7f9b/urdf/robots/rexrov_description/urdf/rexrov_robot.urdf.xacro#L13

**Comment by mabelzhang** [comment]
Oh that package is in `uuv_manipulators`. I need to update and retry.

**Comment by j-herman** [comment]
@crvogt The demo comes up and runs fine (on my local system - haven't tried again with docker) but the arms didn't return to the start pose - both made it to the ee pose, and then I got "Unable to assume pose" in the bimanual_simple_demo.py window.  The arms were still controllable via MoveIt.  I'm going to take a look and see what that message is doing, and run it again.  It's not unusual for MoveIt commands to just fail sometimes, so the demo may be fine.

**Review by mabelzhang — COMMENTED** [review]
Dependencies, for anyone trying to run this:
- Latest version from https://github.com/Field-Robotics-Lab/uuv_manipulators
- Install additional binary packages for controllers and MoveIt https://github.com/Field-Robotics-Lab/dockwater/pull/21

Once merged, should update this wiki page:
https://github.com/Field-Robotics-Lab/dave/wiki/Install-Directly-on-Host
The `sudo apt install` line and maybe the "Note" line below it.

Test result:
After closing the gripper sand moving to the first goal poses, it prints a message and doesn't move on to the second goal pose. I killed all 3 nodes and reran them 3 times with the same result in dockwater:
```
==========Opening right gripper...==========

==========Closing right gripper...==========

==========Opening left gripper...==========

==========Closing left gripper...==========

==========Sending right ee to goal...==========

==========Sending left ee to goal...==========
Unable to assume pose
```
Any ideas?

Other suggestions:
- The Python node is not responding to keyboard interruption (Ctrl+C), so it keeps running perpetually, and I couldn't kill it when it got stuck after Gazebo died. There were many `python3` processes in `top`, so it wasn't easy to find the right process to kill. I just did Ctrl+Z and left it suspended.

   I see that it listens for keyboard interrupt in `main`, but it's getting stuck somewhere else. One case was somewhere in `run_nodes`, and another was before `rosmaster` was started. There probably needs to be more try-except blocks added, or maybe some of the calls to MoveIt need timeouts.

- Currently, when Gazebo starts up, it's looking at some sand and no robot. It would be better to set the SDF world `<gui><camera><pose>` (like [here](https://github.com/Field-Robotics-Lab/dave/pull/186/files)) to point at the robot so that it's visible at startup. For now, the user can go into the Gazebo GUI World list on the left > Models > right-click rexrov > Move To, to get to the robot.

**Comment by bsb808** [comment]
@crvogt It sounds like this PR would benefit from resolving the dockwater PR
- Install additional binary packages for controllers and MoveIt https://github.com/Field-Robotics-Lab/dockwater/pull/21

Can you please address the dockwater PR as well so that things stay (somewhat) in sync?

**Comment by crvogt** [comment]
@mabelzhang @j-herman 
`Unable to assume pose` just means it couldn't find a plan. I'll change the target goals and check that it can be exited if it doesn't reach a pose.

@mabelzhang Regarding the world setup, since multiple examples use that world, I was trying to avoid changing it. I can change the spawn location of the ROV though.

@bsb808 It's not a docker/dependency issue, but that PR can be merged.

**Comment by j-herman** [comment]
Same results on repeat as Mabel is getting - possibly an issue with how the start pose is being stored?

**Comment by crvogt** [comment]
> Same results on repeat as Mabel is getting - possibly an issue with how the start pose is being stored?

Could be, I'm checking it now.

**Comment by crvogt** [comment]
@mabelzhang @j-herman 
Turned out to be two things: Python environment setting was wrong = hanging, and leftover variable = execution failure.

**Comment by j-herman** [comment]
Everything looks good!

**Comment by mabelzhang** [comment]
Camera pose much better! Can move to all poses.

Still cannot kill with Ctrl+C once the routine has started though:
```
==========Opening right gripper...==========
^C
==========Closing right gripper...==========
^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C
==========Opening left gripper...==========

==========Closing left gripper...==========
^C^C
==========Sending right ee to goal...==========
^C^C^C^C^C
```

**Comment by crvogt** [comment]
> Camera pose much better! Can move to all poses.
> 
> Still cannot kill with Ctrl+C once the routine has started though:


Ah, I see. I hadn't tried to stop it.

**Review by mabelzhang — APPROVED** [review]
Keyboard interrupt can be in a followup PR. I do think that should be fixed, because the routine takes like 5 minutes, and the user should have an option of shutting it down early. I ticketed an issue to track it outside this PR https://github.com/Field-Robotics-Lab/dave/issues/230

Reminder to update the wiki page with the new dependencies after merge.
https://github.com/Field-Robotics-Lab/dave/wiki/Install-Directly-on-Host

**Comment by crvogt** [comment]
> Keyboard interrupt can be in a followup PR. I do think that should be fixed, because the routine takes like 5 minutes, and the user should have an option of shutting it down early. I ticketed an issue to track it outside this PR #230

It doesn't appear possible to interrupt the movements. I instead nested it in a for loop with "action numbers". It can at least now be interrupted between actions.

**Comment by mabelzhang** [comment]
Yeah, I think that suffices. I was more referring to interrupting between the actions, since they are different functions in the Python script, and we have full control over them.

**Comment by mabelzhang** [comment]
I'll make a followup PR in 15 minutes or so.

**Comment by crvogt** [comment]
Do you mean interrupt while it's executing? I'm not having trouble interrupting between `===...===` lines.
```
==========Opening right gripper...==========
^Croot@b853c34f61e1:/home/carson/dave_ws
```

mabelzhang:
I retried with `master` branch and it works now! I guess I didn't pull the latest after the PR was merged. All good! Thanks!

crvogt:
I was going to say I think you approved the PR as I made a commit, but I was 13 min behind you. Sorry for the sabotage :/

Issue #224, opened 2022-02-25
Removing dave from dave_sim.repos


bsb808:
At this point this is just to link to the comment in the merged PR so that it doesn't fall through the cracks.  @M1chaelM is probably best positioned to describe what needs to be done post 4.3.0 release.

See: https://github.com/Field-Robotics-Lab/dave/pull/199

> This looks good, and we're going to try it out. It sounds like removing the dave repos from dave_sim.repos would ultimately make sense also, though we will do that later just for logistical reasons relating to our upcoming release.

M1chaelM:
This is an optional optimization, so I was going to ask in our meeting whether we want to do it.

The situation is that when `vcs` was set up, the person who created the `dave_sim.repos` file inlcuded the dave repo in the file. This leads to a little bit of redundancy because you have to retrieve the dave repo to get the `dave_sim.repos` file in the first place. Then when you use `vcs` to get the rest of the repos for the first time, you have to ignore the fact that you downloaded the `dave` repo already. After the first time, it's very easy to use vcs to update all the repos at once, including the dave repo.

We've been living with this tiny bootstrapping requirement and using the `--skip-existing` flag to avoid getting an error the first time we run `vcs` to clone all the repositories. Since we are usually not wiping out our workspace and starting from scratch, this is not a big problem. 

Now that we are moving to a more sophisticated CI script that is going to rely on our `dave_sim.repos` file, having the dave repository in it makes the CI take a bit longer. The reason is that the CI script process is to automatically build and test the main repo (dave) and then build and test everything in the `dave_sim.repos` file. Since that file currently includes the dave repo itself, we end up testing it twice.

All of this is just a restatement of what our kind contributor pointed out when submitting the PR to add `industrial_ci`. There are two ways to avoid this redundancy and make the CI faster:

1. Remove the dave repo from `dave_sim.repos`. Our contributor says this is "preferred." If that's the case then it means we have probably not been using `vcs` in the best way, and there is some better workflow. This seems plausible but I'm not yet sure what that workflow is, so we need to learn a little more about `vcs`. So this change would affect our current workflow and installation instructions.
[...]

PR #229, opened 2022-02-25
removing dave from `dave_sim.repos`

M1chaelM:
This small tweak addresses issue #224 by removing the entry for the dave repo from `dave_sim.repos`, which should really be a list of repos that we need **in addition** to dave.

To test:
```
mkdir -p ~/test_ws/src
cd ~/test_ws/src
git clone https://github.com/Field-Robotics-Lab/dave.git
cd ~/test_ws/src
vcs import --input dave/extras/repos/dave_sim.repos
```
This should import all repos needed for dave. Next run:
```
vcs status
```
And notice that you get a status of all the repos in your `src` directory, including dave.

This change does not affect functionality and is not extremely pressing. Once this PR is ready to merge we should update the installation instructions on the Wiki: https://github.com/Field-Robotics-Lab/dave/wiki/Clone-Dave-Repositories

That's also not too pressing though because the current instructions will still work. They just include some unnecessary characters.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
 
     env:
       UPSTREAM_WORKSPACE: extras/repos/dave_sim.repos -ds_sim
-      TARGET_WORKSPACE: . github:ds_sim#nps_dev
+      TARGET_WORKSPACE: . github:Field-Robotics-Lab/ds_sim#nps_dev
       BASEDIR: ${{ github.workspace }}/.work
       AFTER_SETUP_UPSTREAM_WORKSPACE: vcs pull $BASEDIR/upstream_ws/src
       DOCKER_IMAGE: woensugchoi/dave:latest


bsb808:
@M1chaelM Let's knock this out if you have time.  I can do the review if you can edit the wiki: installation instructions on the Wiki: https://github.com/Field-Robotics-Lab/dave/wiki/Clone-Dave-Repositories

bsb808:
@M1chaelM The fix looks good, but CI is failing.  I know it is interrelated, do you expect this PR to fail the current CI?  If so, happy to approve.

M1chaelM:
> @M1chaelM The fix looks good, but CI is failing. I know it is interrelated, do you expect this PR to fail the current CI? If so, happy to approve.

@bsb808 No, this is not expected! I don't know why it's failing. For some reason, all my headaches seem to come from the  `ds_sim` repo.

M1chaelM:
OK, so it looks like what's happening is that the `ds_sim` package depends on the `dave_gazebo_world_plugins` package, but since dave is no longer contained in `dave_sim.repos` this package is not built until _after_ all the upstream packages are built. So in other words, we're treating `ds_sim` as upstream but it's not really upstream, and it should not be built until dave is built. I'm not too familiar with these tools, so I need to figure out how to address this.

@senceryazici Did you encounter this problem when you removed dave from the `dave_sim.repos` file?

bsb808:
@M1chaelM Sounds like we should probably delay until after 4.3.0 - true?  I underestimated the complexity yesterday when I proposed we do this as part of today's release.

I'll move to milestone 4.3.1 until I hear otherwise.

M1chaelM:
> @M1chaelM Sounds like we should probably delay until after 4.3.0 - true? I underestimated the complexity yesterday when I proposed we do this as part of today's release.
> 
> I'll move to milestone 4.3.1 until I hear otherwise.

Yes, I thought this would be easy too. I agree with moving to 4.3.1.

dtdavi1:
@M1chaelM  The cross dependency is because of the interplay between the ocean current plugin in dave and the DVL plugin in ds_sim.  I don't recall the specifics, but we did talk about it before (probably around release 2.0).  We can revisit it next week if it's causing a problem.

bsb808 (approved):
good to go

woensug-choi:
Merging it as it seems final answer was 'good to go'

PR #231, opened 2022-02-28
Teleporting the vehicle to different demo stations in the integrated world

quarkytale:
Load the integrated world 
```
roslaunch dave_demo_launch dave_integrated_demo.launch
```
and run the script in another terminal.
```
python3 examples/dave_nodes/src/teleport_vehicle.py integrated_world_demo_stations.yaml
```
Input `<vehicle> <location>: rexrov home` to move to the vehicle to the desired pose and orientation.

Note: I haven't added exact locations since the demo stations' locations isn't final yet, will just need to update the yaml file.


examples/dave_nodes/src/teleport_vehicle_demo_stations.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+'''
+Teleport the vehicle to different demo stations
+
+'''
+
+# System imports
+import math
```
Yadunund, line 8: unused import

Yadunund (changes requested):
Thanks for working on this! I've left some feedback in the comments. Overall I think it would be great if we can architecture this script to be reusable for different worlds. One way we can do this is to have the script parse a yaml file via argparse. This yaml file can have a schema as below

```yaml
robot_name:
  places: [{place_name: pose}]
```

So an example of this would be
```
rexrov1:
  places: [
    {'home': [-21,-34,-108,0,0,2.5]},
    {'electrical': [-30,-34,-108,0,0,3.141592]}
  ]
```

Then inside `main()` receive the path to this file via `argparse`, parse this yaml and save a dictionary mapping robot names to a dictionary of place names and poses. 

I think it would be nice if inside main we run a `while not rospy.is_shutdown()` inside which we wait for users to input a robot_name and place name to teleport to. (We could printout the list of places in the config as a hint). Then every time, there is a valid input, we call `teleport(robot_name, place_name)` which waits for the gazebo service to be available and then publishes the request to update the robot model's state. Once done, the script waits for the next input from the user.

This way this script can be reused for multiple worlds. Users would just create separate yaml config file. Let me know if you have any questions.

quarkytale:
Thanks for helping out! Updated the script as per your feedback, let me know if anything could be further improved.

Yadunund (approved):
There were a couple of issues with the script including hardcoded path to config file which depended on the working directory from which the script was run. I've fixed this and other issues. Users can now run the script with 
```
rosrun dave_nodes teleport_vehicle.py -c CONFIG_PATH
```

PR #235, opened 2022-03-10
Add dependency for moveit_commander

Yadunund:
When trying out the bimanual PR https://github.com/Field-Robotics-Lab/dave/pull/228 with a workspace I setup following instructions from the wiki, I get a `import moveit_commander ModuleNotFoundError: No module named 'moveit_commander'` error.  when running `rosrun dave_nodes bimanual_simple_demo.py`

This PR adds `moveit_commander` package as a dependency to `dave_nodes` and as a result will be installed when `rosdep` is run via the command `rosdep install --from-paths src --ignore-src --rosdistro noetic -y` 

Note: The above `rosdep` commands installs the other moveit packages that are required for the demo. Only the `moveit_commander` package was missing.

### DIFF
--- a/examples/dave_nodes/package.xml
+++ b/examples/dave_nodes/package.xml
@@ -7,6 +7,9 @@
   <license>Apache2</license>
   <author email="dtdavi1@nps.edu">dtd</author> -->
   <buildtool_depend>catkin</buildtool_depend>
+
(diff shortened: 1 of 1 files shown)


crvogt (approved):
Thanks for the addition and trying the example.

PR #241, opened 2022-03-18
Bimanual manipulation update

crvogt:
Update to PR #228 

Depends on Field-Robotics-Lab/uuv_manipulators#9

Expected scenario:
Demonstrates grasping and stability while holding a static object and picking up a pot.
<p align="center">

</p>

Changes made:
- Added controller arguments (effort vs position)
- Added arguments for ROV initial position in launch file

To test
```
roslaunch dave_demo_launch dave_oberon7_moveit.launch
```
After allowing time for startup, run
```
roslaunch rexrov_oberon7_moveit rexrov_dual_arm_moveit_planning_execution.launch moveit_controller_manager:=rexrov
```
Once planning can begin, run the updated simple demo
```
rosrun_dave_nodes bimanual_simple_demo.py
```

Notes:
- Occasionally the pot slips out of the gripper, I adjusted the pot friction values locally.

### DIFF
--- a/examples/dave_demo_launch/launch/dave_oberon7_moveit.launch
+++ b/examples/dave_demo_launch/launch/dave_oberon7_moveit.launch
@@ -4,8 +4,7 @@
 	<arg name="joy_id" default="0"/>
 	<arg name="paused" default="false"/>
 	<arg name="gui" default="true"/>
-	<arg name="effort_controller" default="true"/>
-	<arg name="position_controller" default="false"/>
+	<arg name="controller_type" default="effort"/>
 
 	<include file="$(find gazebo_ros)/launch/empty_world.launch">
 		<arg name="world_name" value="$(arg world_name)"/>
@@ -22,8 +21,7 @@
 
     <!-- Upload URDF with correct transmission type -->
 	<include file="$(find rexrov_description)/launch/upload_rexrov_oberon7_moveit.launch">
-		<arg name="controller_type" value="effort" if="$(arg effort_controller)"/>
-		<arg name="controller_type" value="position" if="$(arg position_controller)"/>
+		<arg name="controller_type" value="$(arg controller_type)"/>
 		<arg name="x" value="2.5"/>
 		<arg name="y" value="0.5"/>
 		<arg name="z" value="-97.5"/>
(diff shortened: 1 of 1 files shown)


mabelzhang:
Ooh this is cool.

Since we're here, could I ask for a favor? Could you throw this extra object into the world:
```
<uri>model://sunken_vase_distorted</uri>
```
as part of the "scenario: degraded object manipulation" task? Our current scenario for that task has a multibeam sensor but not an arm, so it's more perception than manipulation. We can either throw an arm into that world, or throw a distorted object in a world that already has an arm. This world looks perfect with its existing stock of 1 vase. We could make it 2 vases.

Thanks in advance!

crvogt:
@mabelzhang Do you want me to do anything else with it? Ie, set it as a target to pick up or anything?

examples/dave_demo_launch/launch/dave_oberon7_moveit.launch
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0"?>
 <launch>
-	<!-- <arg name="world_name" default="worlds/empty.world"/> -->
-	<arg name="world_name" default="$(find dave_worlds)/worlds/dave_ocean_waves.world"/>
+	<arg name="world_name" default="$(find dave_worlds)/worlds/dave_bimanual_example.world"/>
 	<arg name="joy_id" default="0"/>
 	<arg name="paused" default="false"/>
 	<arg name="gui" default="true"/>
+	<arg name="effort_controller" default="true"/>
```
Yadunund, line 7: By having two separate boolean variables to select controllers, this launch file will try to end up spawning the same controller manager twice which will lead to an exception being thrown if the users launches the file with `position_controller:=true` and not explicitly passing `effort_controller:=false`. See below.

```
yadu@yadu:~/ws_dave$ roslaunch dave_demo_launch dave_oberon7_moveit.launch position_controller:=true
... logging to /home/yadu/.ros/log/2745cdaa-aa96-11ec-883e-e9d7e55577dc/roslaunch-yadu-112151.log
Checking log directory for disk usage. This may take a while.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

RLException: Invalid <arg> tag: arg 'controller_type' has already been declared. 

Arg xml is <arg name="controller_type" value="position" if="$(arg position_controller)"/>
The traceback for the exception was written to the log file
```

One way to avoid this problem (and I think you're already doing this in `upload_rexrov_oberon7_moveit.launch`), is to have a single `controller_type` argument here with default value `effort`. We can directly pass this argument when we include `upload_rexrov_oberon7_moveit.launch`. The `goup if` block below where we spawn the controller_manager can also just check for this string value.

Yadunund (changes requested):
Tested and works well. The vase does drop at times but I think fixing that will require some contact tuning and potentially simplifying the mesh for the vase handle. 

But apart from that I have a small recommendation for the launch file. Happy to merge in after that is addressed.

examples/dave_demo_launch/launch/dave_oberon7_moveit.launch
crvogt: @Yadunund Thank you for the catch. I've modified the file to accept a single `controller_type` arg as per your suggestion.

Yadunund (approved):
Thanks for addressing the feedback!

PR #243, opened 2022-03-22
Add Doc Ricketts ROV 

crvogt:
Test with `roslaunch dave_robot_launch doc_ricketts.launch`


bsb808:
@crvogt Unless MBARI signs off on using the name "Doc Ricketts" - we will want to use a pseudonym.

PR #250, opened 2022-03-25
Change exec_depend to run_depend

Yadunund:
PR https://github.com/Field-Robotics-Lab/dave/pull/239 moved a couple exec dependencies from `dave_nodes` pkg into `rexrov_oberon7_moveit`. However, the former's `package.xml` is `v3.1.1` which supports `exec_depend` while the latter is still on `v0.3.0` which does not.

Fix https://github.com/Field-Robotics-Lab/dave/issues/248

Signed-off-by: Yadunund <yadunund@openrobotics.org>

PR #239, opened 2022-03-16
Add another moveit dependency

Yadunund:
I missed another dependency, `moveit-planners`, in PR https://github.com/Field-Robotics-Lab/dave/pull/235

Without this, launching the controller manager has some errors displayed. 
```
[ERROR] [1647354067.549891762]: Exception while loading planner 'chomp_interface/CHOMPPlanner': According to the loaded plugin descriptions the class chomp_interface/CHOMPPlanner with base class type planning_interface::PlannerManager does not exist. Declared types are  ompl_interface/OMPLPlanner
Available plugins: ompl_interface/OMPLPlanner
```

```
[ERROR] [1647354067.556044888]: Failed to initialize planning pipeline 'chomp'.
```

```
[ERROR] [1647354067.615621238, 37.172000000]: Exception while loading planner 'pilz_industrial_motion_planner::CommandPlanner': According to the loaded plugin descriptions the class pilz_industrial_motion_planner::CommandPlanner with base class type planning_interface::PlannerManager does not exist. Declared types are  ompl_interface/OMPLPlanner
Available plugins: ompl_interface/OMPLPlanner
[ERROR] [1647354067.615665542, 37.172000000]: Failed to initialize planning pipeline 'pilz_industrial_motion_planner'.
```

Signed-off-by: Yadunund <yadunund@openrobotics.org>

PR #251, opened 2022-03-25
Checking for quick CI fix

j-herman:
@mabelzhang This gets us past the CI issue but may not be correct - over to you whether you'd like to use it.  If not, feel free to delete.


mabelzhang (approved):
Ohhhh. You and I were focusing on two different package.xml's.
Yeah, I think this is definitely needed here, because `examples/dave_nodes/src/bimanual_simple_demo.py` has `import moveit_commander`.
I think #239 moved it to `urdf/robots/rexrov_oberon7_moveit/package.xml`, but it is still needed in `dave_nodes`.

This is related to #239 and #250.

PR #253, opened 2022-03-25
Replace moveit_planners with moveit_planners_chomp

mabelzhang:
Trying to fix CI broken in #239 and #250
See https://github.com/Field-Robotics-Lab/dave/pull/250#issuecomment-1079502754 and subsequent comments.

CI might be failing because `moveit_planners` doesn't exist in the Docker image. These ones exist though: `moveit_planners_ompl` (already a dependency) and `moveit_planners_chomp`.
Replace `moveit_planners` with `moveit_planners_chomp` to see if that fixes CI.

Rearrange in alphabetical order while I'm at it.


j-herman:
Tested with the dual arm rviz demo and everything still runs fine.  If this fixes CI I'll approve for you.  It does make me wonder, though, why the other fix worked when moveit_planners was still included in the file.

j-herman (approved):
Verified it all still works with the bimanual manipulation demo

mabelzhang:
I don't know what the CI is checking, but seems like it's doing more than our manual builds.
It could be that our manual builds aren't running into problems because it's a runtime dependency. Then at runtime, maybe the _ompl package is enough, and _chomp isn't really needed. I don't know.
Either way, [moveit_planners](https://github.com/ros-planning/moveit/blob/master/moveit_planners/moveit_planners/package.xml) seems to be a meta package that has runtime dependency on both _ompl and _chomp, so depending on that should work, except we don't have that package in our Docker image for some reason.

PR #269, opened 2023-05-04
Update repo name and location for acoustic messages

rolker:
This addresses #268.

### DIFF
--- a/extras/repos/multibeam_sim.repos
+++ b/extras/repos/multibeam_sim.repos
@@ -6,4 +6,4 @@ repositories:
   marine_msgs:
     type: git
     url: git@github.com:apl-ocean-engineering/marine_msgs.git
-    version: main
+    version: tags/v2.0.0


lauralindzey:
In order to future-proof this, you may want to use a specific tag or commit for the marine_msgs repo, rather than having it use `main`.

mabelzhang:
Might fix errors like this one on `acoustic_msgs` https://github.com/Field-Robotics-Lab/dave/issues/272
But the package names in hydrographic_msgs have also been changed, right? So `acoustic_msgs` are no longer that, it has to be changed to [`marine_acoustic_msgs`](https://github.com/apl-ocean-engineering/marine_msgs/tree/main/marine_acoustic_msgs) now?

😅 I got confused in that issue, because now there's another package that's been released and called `acoustic_msgs` 😅 and that one had a genuine error that we ran into a few weeks back. It could get very confusing for somebody.

It `marine_msgs` still for Noetic? I ask because that's what DAVE targets. If you will have different branches, it would also help if we can use the Noetic one here.

I don't have a setup to test it anymore. I don't know if anybody has at this point. So I don't know if we'll be merging PRs.

lauralindzey:
@mabelzhang -- just a note that I'm at sea, and will be getting home late August. I tried (and failed) to get marine_acoustic_msgs properly released before I left, so it's close to the top of my stack for when I get home.

> But the package names in hydrographic_msgs have also been changed, right? So acoustic_msgs are no longer that, it has to be changed to [marine_acoustic_msgs](https://github.com/apl-ocean-engineering/marine_msgs/tree/main/marine_acoustic_msgs) now?

Correct.

woensug-choi (changes requested):
Thank you for the PR! 
Could you change the `version:main` to `tags/v2.0.0`? This way we can fix it to the v2.0.0 tag for future-proof

mabelzhang:
Ooh that's a good idea.

Issue #27, opened 2020-07-23
OpenCV compilation error


PenguineZhang:
Is there opencv version requirement to compile the code? My workaround is to change CV_AA to CV_L1 as suggested by the compiler. My machine is using OpenCV 4.1.0. See attached and let me know what you think. 
![Screenshot from 2020-07-22 21-12-24](https://user-images.githubusercontent.com/17698544/88251585-1d605280-cc60-11ea-933b-49a756c2204d.png)

bsb808:
This does seem like a dependency issue with the setup.  That source file does make use of...
```
#include <cv_bridge/cv_bridge.h>
#include <opencv2/core/core.hpp>
```

Can you make sure you have the setup as described [here](https://github.com/Field-Robotics-Lab/dave/wiki/Install-Directly-on-Host).  One thing that causes issues is to make sure you have the updated Gazebo 9 from OSRF.  It should be `Gazebo multi-robot simulator, version 9.13.1`


@BruceMty Do we need the sonar_dave_plugins anymore?  It look like this is a copy of the source file from the [uuv_simulator](
https://github.com/uuvsimulator/uuv_simulator/blob/master/uuv_sensor_plugins/uuv_sensor_ros_plugins/src/gazebo_ros_image_sonar.cpp)  Can we remove that part of the repository?

BruceMty:
@bsb808 , this is a copy but we currently need it, see https://github.com/Field-Robotics-Lab/dave/issues/25.  I expect that uuv_simulator will raise the same CV complaint.  For me, typing "pkg-config --modversion opencv" indicates CV version 3.2.0.

PenguineZhang:
@BruceMty @bsb808 Upgrading to Gazebo 9.13.2 (should be the latest) still raised the same compilation issue. The output of "pkg-config --modversion opencv" also indicates 3.2.0. It turned out CV_AA is defined in "opencv2/imgproc/imgproc_c.h". When I added include <opencv2/imgproc/imgproc_c.h>, it compiled sucdessfully.

bsb808:
That is strange that we aren't seeing the same thing - suggests that the development environment is not equivalent.  This is a perfect reason why we need some CI/CD to provide checks for this in a clean dev. env. 

I suggest we incur the technical debt and use this as our first test case of @PenguineZhang submitting a PR through gitlab that eventually makes it to the github repo.

PenguineZhang:
[...]

Issue #274, opened 2023-07-14
Integrated Demo: demo world not launching


mjsaenz:
When I follow the instructions for the [integrated demo](https://field-robotics-lab.github.io/dave.doc/contents/integrated_world), I am unable to get the demo world to load. The terminal will fill with the error seen below, while Gazebo will take a long time to launch. Once it does launch, however, the world screen remains black. Is this due to an underpowered machine?

![image](https://github.com/Field-Robotics-Lab/dave/assets/75288078/119b6df4-c1ba-485f-964c-e1869a39f973)
_The aforementioned error._

PR #282, opened 2023-12-01
`fuel.ignition.org` to `fuel.gazebosim.org`

woensug-choi:
As Ignition is now Gazebo (the name we all love). The models loading from `Fuel Ignition` (still old name) url is changed.
The models with loading from the `Fuel Ignition`, the cloud model library, need url changed.

This should fix https://github.com/Field-Robotics-Lab/dave/issues/27 and https://github.com/Field-Robotics-Lab/dave/issues/274

I will add some text in `dave.doc` for loading time (in my system, it took almost 5 minutes to download all models)

### DIFF
--- a/examples/dave_demo_launch/launch/dave_integrated_demo.launch
+++ b/examples/dave_demo_launch/launch/dave_integrated_demo.launch
@@ -7,7 +7,7 @@
     <arg name="set_timeout" default="false"/>
     <arg name="timeout" default="0.0"/>
     <arg name="velocity_control" default="true"/>
-    <!-- <arg name="joy_id" default="0"/> -->
+    <arg name="joy_id" default="0"/>
     <arg name="debug" default="false"/>
 	<!-- <arg name="controller_type" default="effort"/> -->
 


mabelzhang:
Added to WG kanban https://github.com/orgs/ros-maritime/projects/1

examples/dave_demo_launch/launch/dave_integrated_demo.launch
```diff
@@ -7,12 +7,10 @@
     <arg name="set_timeout" default="false"/>
     <arg name="timeout" default="0.0"/>
     <arg name="velocity_control" default="true"/>
-    <arg name="joy_id" default="0"/>
+    <!-- <arg name="joy_id" default="0"/> -->
```
mabelzhang: Are these commented tags supposed to be added to this PR, or is it a mistake? I didn't see a note on these

woensug-choi: I will double check whether this was needed or mistake soon. Thanks!

mabelzhang:
Hmm, I see that these are the new models we added in 2021. We have a copy of these on Fuel, those might need to be re-uploaded as well... If the URIs are actually pulling from Fuel and not the ones in this local repo, fixing the files in this repo wouldn't fix the error. But I see in the URI that the models on Fuel are all owned by different people... I don't think I can replace them. They would all have to be re-uploaded to a different account.

I haven't test it. Does this PR get rid of the errors for you?

If this PR doesn't fix the error, then maybe we should just refer to the local copies of the models and not use the ones on Fuel... until they can be re-uploaded. Ughhhh. But that's undoing all the uploading we did. And porting to new Gazebo actually works better with models on Fuel.

Sigh, the Ignition rename really sucks. We just had the same kind of error cropped up in the splash screen worlds in Harmonic today.

woensug-choi:
Oh wow.. I haven't tested everything but the ones I've tested worked. This changes were first brought up after issues that the world was not launching. Modifying the URL as in this PR fixed it.

woensug-choi:
@mabelzhang I shall be done with your review soon.. sorry. What do you think of publishing a new Release after this PR as it fixes the launch of world with new URL.

Added Uncommenting commit for joystick. This should be good to go.

examples/dave_demo_launch/launch/dave_integrated_demo.launch
woensug-choi: Added Uncommenting commit. This should be good to go.

mabelzhang:
Oh I meant in the diff for `examples/dave_demo_launch/launch/dave_integrated_demo.launch`, there's still a lot of other commented tags. Do they all need to be commented? The joystick was just one of many.

